PR #62, opened 2021-02-24
don't fail on duplicate likes and increase demo feed

ruslandoga:


ruslandoga:
superceeded by https://github.com/ruslandoga/test3/pull/63

PR #281, opened 2021-09-10
ready?: fix current call return as missed

akhrail1996:


ruslandoga (changes requested):
can't compare with nil

akhrail1996:
@ruslandoga 
what's the proper way?

lib/t/calls.ex
```diff
@@ -153,6 +153,7 @@ defmodule T.Calls do
     Call
     |> where(called_id: ^user_id)
     |> where([c], is_nil(c.accepted_at))
+    |> where([c], c.ended_at != nil)
```
ruslandoga, line 156: `not is_nil(c.ended_at)`

PR #283, opened 2021-09-13
wip: deactivated session push notification

akhrail1996:


lib/t/feeds.ex
```diff
@@ -45,9 +45,15 @@ defmodule T.Feeds do
 
     for user_id <- user_ids do
       broadcast(topic, {__MODULE__, :deactivated, user_id})
+      schedule_deactivated_push_notification(user_id)
```
ruslandoga, line 48: `notify_deactivated` are only for in-app broadcasts

lib/t/feeds.ex
```diff
@@ -45,9 +45,15 @@ defmodule T.Feeds do
...
     end
   end
 
+  defp schedule_deactivated_push_notification(user_id) do
+    job = DispatchJob.new(%{"type" => "deactivated", "user_id" => user_id})
+    Oban.insert(job)
```
ruslandoga, line 54: this would insert each job separately -- 100 users -> 100 inserts. It needs to be done in one insert

lib/t/push_notifications/dispatch_job.ex
```diff
@@ -126,6 +126,16 @@ defmodule T.PushNotifications.DispatchJob do
     |> Repo.one()
   end
 
+  defp handle_type("deactivated" = type, args) do
+    %{"user_id" => user_id} = args
```
ruslandoga, line 130: see the warning above

lib/t/feeds.ex
akhrail1996: yeah it was. does adding push_scheduling somehow breaks it / slows it down?

lib/t/push_notifications/dispatch_job.ex
akhrail1996: yeah, saw it but couldn't understand since it point at L14:
defp handle_type("match", args) do
    %{"match_id" => match_id} = args

however we also have similar guys like:
1) defp handle_type("timeslot_cancelled" = type, args) do
    %{"match_id" => match_id} = args
or
2) defp handle_type("timeslot_ended", args) do
    %{"match_id" => match_id} = args

which also seem to have the same name and arity (number of arguments), but do not produce this kind of warning

what am I missing here?

lib/t/feeds.ex
ruslandoga: yes, it needs to stay the way it was

akhrail1996: could you please explain the logic?  (I'd like to understand who it works)

ruslandoga: that's just how it works right now

ruslandoga: I guess it's usually called separation of concerns

lib/t/feeds.ex
ruslandoga: for example, because you are using the wrong place to schedule the notifications, you can't optimise this insert

lib/t/push_notifications/dispatch_job.ex
ruslandoga: > clauses with the same name and arity (number of arguments) should be grouped together

you need to put the new clause next to the other ones, above `profile_name/1`

ruslandoga:
https://github.com/ruslandoga/test3/pull/284

PR #296, opened 2021-09-24
add ended_by to calls

ruslandoga:


lib/t/calls.ex
```diff
@@ -147,12 +147,14 @@ defmodule T.Calls do
...
   end
 
-  # TODO remove the ones hung up by user_id
   @spec missed_calls_q(Ecto.UUID.t(), Keyword.t()) :: Ecto.Query.t()
   defp missed_calls_q(user_id, opts) do
     Call
     |> where(called_id: ^user_id)
+    # call hasn't been picked up
     |> where([c], is_nil(c.accepted_at))
+    # and call hasn't been declined
```
akhrail1996, line 156: declined call is not exactly "missed", but still needs to be displayed for ability to conveniently call back

ruslandoga: @akhrail1996 if a call has been declined, it should be stored in ios already though, right?

ruslandoga: `missed calls` are supposed to only be the ones ios might not know about.

ruslandoga: If however you consider users having multiple devices, we'd need to return all calls, even successful ones, on feed channel join.

akhrail1996: declined calls are not stored in local db yet, though they are inserted in feed
therefore on app refresh they will be lost, which is bad

yeah, probably returning all calls will work best

ruslandoga: https://github.com/getsince/test3/issues/298

PR #312, opened 2021-10-04
call's notification

orlovvlad95:


lib/t/calls.ex
```diff
@@ -104,6 +107,9 @@ defmodule T.Calls do
 
   @spec end_call(Ecto.UUID.t(), Ecto.UUID.t(), DateTime.t()) :: :ok
   def end_call(user_id, call_id, now \\ utc_now()) do
+    m = "user #{user_id} ended a call #{call_id}"
+    Logger.warn(m)
+    Bot.async_post_message(m)
```
ruslandoga, line 112: @orlovvlad95 CI would fail here, not sure why it didn't run for your PR ...

ruslandoga: Ci would say that `calls.ex` is not formatted because a newline is missing.

ruslandoga (approved):

PR #331, opened 2021-10-15
ready: add call topics

akhrail1996:


config/runtime.exs
```diff
@@ -197,8 +197,11 @@ if config_env() == :dev do
     salt: System.fetch_env!("IMGPROXY_SALT")
 
   config :t, T.Bot,
-    token: System.fetch_env!("TG_BOT_KEY"),
-    room_id: System.fetch_env!("TG_ROOM_ID") |> String.to_integer()
+    token: "asdfasdfasdf",
```
ruslandoga, line 200: just set the env vars

lib/t/application.ex
```diff
@@ -23,7 +23,8 @@ defmodule T.Application do
         TWeb.Telemetry,
         maybe_migrator(),
         {Oban, oban_config()},
-        Supervisor.child_spec({Task, &T.Release.mark_ready/0}, id: :readiness_notifier)
+        Supervisor.child_spec({Task, &T.Release.mark_ready/0}, id: :readiness_notifier),
+        T.CallTopics
```
ruslandoga, line 27: needs to be started before endpoint

lib/t/call_topics.ex
```diff
@@ -0,0 +1,34 @@
...
+    locales = ["ru", "en"]
+
+    for locale <- locales do
+      {:ok, raw_topics} = File.read("priv/call_topics/#{locale}.txt")
+      topics = String.split(raw_topics, "\n", trim: true)
+
+      :ets.insert(call_topics_pid, {locale, topics})
+    end
+
+    {:ok, call_topics_pid}
```
ruslandoga, line 22: table is not a pid

lib/t/call_topics.ex
```diff
@@ -0,0 +1,34 @@
...
+  use GenServer
+
+  @table_name :call_topics
+
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, nil, name: __MODULE__)
+  end
+
+  def init(_init_state) do
+    call_topics_pid = :ets.new(@table_name, [:set, read_concurrency: true])
```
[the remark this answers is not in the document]
ruslandoga: no need for this process as well

lib/t/call_topics.ex
```diff
@@ -0,0 +1,34 @@
...
+
+    locales = ["ru", "en"]
+
+    for locale <- locales do
+      {:ok, raw_topics} = File.read("priv/call_topics/#{locale}.txt")
```
ruslandoga, line 16: just read the file into memory during compilation

test/t/calls/call_topics_test.exs
```diff
@@ -0,0 +1,19 @@
...
+
+  describe "call topics" do
+    # TODO list of supported locales to ENVs?
+    test "call topics for every locale has equal size" do
+      locales = ["en", "ru"]
+
+      en_locale_length = length(CallTopics.locale_topics("en"))
+
+      for locale <- locales do
+        assert en_locale_length == length(CallTopics.locale_topics(locale))
```
ruslandoga, line 15: this doesn't test much, only that there are the same number of items in each locale, both can be 0

lib/t/location.ex
```diff
@@ -13,6 +13,10 @@ defmodule T.Location do
       {:ok, %{"location" => %{"latitude" => lat, "longitude" => lon}}} ->
         [lat, lon]
 
+      {:error, :database_unknown} ->
+        Logger.error("location database_unknown}")
+        [55.755516, 37.615040]
+
```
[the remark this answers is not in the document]
ruslandoga: I explicitly didn't handle that error since it's a setup error and should be make known to sentry. I think with the current setup Logger.error would only appear in cloudwatch logs.

PR #342, opened 2021-10-19
mix most_liked and most_recent in feed

akhrail1996:


ruslandoga (changes requested):
ETS tables can't be created per user, you'll quickly run out of them (only 1400 are allowed by default) and crash the vm. If you want to cache feed per user, store it in feed channel process.

lib/t/feeds.ex
```diff
@@ -97,42 +97,110 @@ defmodule T.Feeds do
...
+      feed_profiles_with_cursor
+      |> Enum.map(fn {_cursor, profile} ->
+        profile
+      end)
+
     {feed_profiles, feed_cursor}
   end
 
+  def save_eligible_feed_profiles_to_memory(user_id, location, gender, gender_preferences) do
+    table = String.to_atom("feed_#{user_id}")
```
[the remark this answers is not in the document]
ruslandoga: https://github.com/getsince/test3/blob/eec09d31d6fa0c55e1780ee17f8917ba080911e3/lib/t/feeds/feed_cache.ex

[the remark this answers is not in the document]
ruslandoga: 1. you need to use ets as a shared resource, meaning reusing the same tables for all users, otherwise just use maps or lists, it'd be more efficient 
2. you need a process controlling the ets table which lives for as long as the app, that process owns the tables https://github.com/getsince/test3/blob/bb7811f56c826cfe1ba79d677086b13339ed2164/lib/t/feeds/feed_cache.ex#L16-L26
3. you split the tables into indexes and data, index tables need to be a tree https://github.com/getsince/test3/blob/bb7811f56c826cfe1ba79d677086b13339ed2164/lib/t/feeds/feed_cache.ex#L18, data tables can be hash tables https://github.com/getsince/test3/blob/bb7811f56c826cfe1ba79d677086b13339ed2164/lib/t/feeds/feed_cache.ex#L22-L23

lib/t/feeds.ex
```diff
@@ -97,42 +97,110 @@ defmodule T.Feeds do
...
+
+    case :ets.whereis(table) do
+      :undefined -> :ok
+      _ -> :ets.delete(table)
+    end
+
+    :ets.new(table, [:ordered_set, :named_table, read_concurrency: true])
```
ruslandoga, line 132: `read_concurrency: true` is only needed when the table is being read from different processes. In current implementation a single feed channel process is reading from the table. Also the behaviour is undefined when two processes read from the same table.

lib/t/feeds.ex
```diff
@@ -97,42 +97,110 @@ defmodule T.Feeds do
...
           {[feed_profile], feed_cursor}
   def fetch_feed(user_id, location, gender, gender_preferences, count, feed_cursor) do
-    feed_profiles =
-      feed_profiles_q(user_id, gender, gender_preferences, feed_cursor)
-      |> limit(^count)
-      |> select([p], {p, distance_km(^location, p.location)})
-      |> Repo.all()
+    case feed_cursor do
+      nil -> save_eligible_feed_profiles_to_memory(user_id, location, gender, gender_preferences)
+      _ -> :ok
```
ruslandoga, line 102: What if there is no ets table?

lib/t/feeds.ex
```diff
@@ -97,42 +97,110 @@ defmodule T.Feeds do
...
+  def feed_profiles_from_memory(user_id, feed_cursor, count) do
+    table_name = String.to_atom("feed_#{user_id}")
+
+    keys = next(table_name, feed_cursor, count)
+
+    feed_profiles =
+      keys
+      |> Enum.map(fn key ->
+        :ets.lookup(table_name, key) |> Enum.at(0)
+      end)
```
ruslandoga, line 170: table is needlessly iterated twice, once in `next` and then again for all keys in `lookup`

PR #356, opened 2021-10-21
don't show hidden profiles

ruslandoga:
@akhrail1996 needs your attention


lib/t/feeds.ex
```diff
@@ -116,7 +116,7 @@ defmodule T.Feeds do
 
   @spec get_mate_feed_profile(Ecto.UUID.t()) :: %FeedProfile{} | nil
   def get_mate_feed_profile(user_id) do
-    FeedProfile
+    not_hidden_profiles_q()
```
ruslandoga, line 119: later `get_mate_feed_profile` can be modified to accept user_id of the requester, then it'd be possible to also filter out reported profiles

PR #379, opened 2021-10-28
expire matches

### LINKED ISSUE #365: expire matches when there is no actions between users
this should incentivise people to connect + declutter their matches

orlovvlad95:
closes #365

### DIFF
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -107,7 +107,6 @@ defmodule T.Factory do
       gender: gender,
       name: opts[:name] || "that",
       times_liked: opts[:times_liked] || 0,
-      last_active: opts[:last_active] || nil,
       filters: %{genders: opts[:accept_genders] || ["F"]}
     }
   end
--- a/test/t_web/channels/feed_channel_test.exs
+++ b/test/t_web/channels/feed_channel_test.exs
@@ -460,11 +460,10 @@ defmodule TWeb.FeedChannelTest do
       assert ed == exp_date
       assert is_binary(match_id)
 
-      import Ecto.Query
+      assert_push "matched", _payload
 
-      event = MatchEvent |> where(match_id: ^match_id) |> T.Repo.one()
-
-      assert %MatchEvent{match_id: ^match_id, event: "created", timestamp: ^now} = event
+      assert %MatchEvent{match_id: ^match_id, event: "created", timestamp: ^now} =
+               Repo.get_by!(MatchEvent, match_id: match_id)
     end
 
     test "when not yet liked by mate", %{


priv/repo/migrations/20211026203244_add_match_events.exs
```diff
@@ -0,0 +1,27 @@
...
+
+    flush()
+
+    match_created_events =
+      "matches"
+      |> select([m], {m.id})
+      |> T.Repo.all()
+      |> Enum.map(fn {match_id} ->
+        %{timestamp: DateTime.utc_now(), match_id: match_id, event: "created"}
+      end)
```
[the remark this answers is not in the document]
orlovvlad95: thanks for you idea 
fixed it

akhrail1996:
- [ ] Disable pruner upon first successful call
- [ ] Add tests

lib/t/matches/match_events.ex
```diff
@@ -0,0 +1,12 @@
+defmodule T.Matches.MatchEvents do
+  @moduledoc false
+  use Ecto.Schema
+
+  @primary_key false
+  @foreign_key_type Ecto.Bigflake.UUID
+  schema "match_events" do
+    field :timestamp, :utc_datetime
+    field :match_id, Ecto.Bigflake.UUID, primary_key: true
```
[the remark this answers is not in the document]
orlovvlad95: fixed

lib/t/matches.ex
```diff
@@ -593,4 +599,26 @@ defmodule T.Matches do
...
+
+  def match_check() do
+    MatchEvents
+    |> distinct([m], m.match_id)
+    |> join(:inner, [m], ma in Match, on: ma.id == m.match_id)
+    |> where([m], m.timestamp < fragment("now() - INTERVAL '48 hours'"))
+    |> select([m, ma], {m.match_id, ma.user_id_1})
+    |> T.Repo.all()
+    |> Enum.map(fn {match_id, user_id} ->
+      T.Matches.unmatch_match(user_id, match_id)
```
akhrail1996, line 611: `unmatch_match` func was made to unmatch one user from another (initiated by other user). here we need a similar function but for expire_match:
* it will do similar stuff in db (delete_match and delete_likes)
* but it will broadcast different event (not `notify_unmatch`, but rather `notify_expired`
* also (minor) - logger message will be different here as well

akhrail1996: @orlovvlad95 on a separate note - I just realised that we will need a separate pruner for "soon to become expired matches" for us to send push-notifications in advance (sth like "your match is about to expire, jump in an schedule an audio date")

akhrail1996:
Actually after doing some work on frontend I am starting to realise that separate `match_events` table may indeed be a suboptimal approach (https://github.com/getsince/test3/issues/365#issuecomment-953675732), however I suggest a bit different approach:
* create a new column `expires_at` in matches table (:datetime)
* upon every relevant match_event this column gets "bumped" (i.e. set to date_now() + matchExpirationDuration interval (48 hours);
* upon successful first call the value becomes `null` - meaning that there is no expiration date for this match no more.

This will allow us:
- to keep eth in one table
- to not calculate match_expiration_date every time (to return to client)
- to prune matches simpler

@ruslandoga @orlovvlad95 thoughts?

akhrail1996:
and to return `expired_match` to every user once (before he/she "seens" it -> from [here](https://github.com/getsince/test5/pull/128#discussion_r740744069)) we can create two bool columns in matches table: `seen_by_user_1` and `seen_by_user_2`

when both users "seen" the expired match -> match and corresponding likes are deleted from db

ruslandoga:
>  I am starting to realise that separate match_events table may indeed be a suboptimal approach, however I suggest a bit different approach:

@akhrail1996 why?

akhrail1996:
@ruslandoga iOS needs to know `match_expiration_date` for every match out there. Seems suboptimal to search `match_events` table for the latest event of every match (+ add `matchExpirationDuration` to arrive at `match_expiration_date`) this user has every time she joins feedChannel.

Just having `match_expiration_date` in matches table seem more natural & simple.

ruslandoga:
it's not suboptimal for db, can be done with one query with a lateral join

akhrail1996:
> having `match_expiration_date` is easy, not simple.

why not simple? and is `match_events` table approach simple then?

ruslandoga:
yes, match_table is simple because it's not dependent on mutable values and only deals with facts. date is not simple because it is mutable, easy to make mistakes (forget to update) and debug (you don't have info on what went wrong since you only have one value, datetime, or its absence = loss of information), and it is also nullable which is the source of many problems both in general programming and working with dbs

akhrail1996:
okay, me resign the `match_expiration_date` approach, thx!

ruslandoga:
I still suggest you try both, it's a good exercise

akhrail1996:
I've tested various scenarios with the client (https://github.com/getsince/test5/pull/128) and so far eth works correctly

@orlovvlad95 and I also plan to add some additional tests, but in general PR is ready for at least high-level review (pretty sure there will be quite a lot of structural comments so probably makes sense to request a review from @ruslandoga now before we jump into details)

lib/t/application.ex
```diff
@@ -26,6 +26,7 @@ defmodule T.Application do
         TWeb.UserSocket.Monitor,
         T.Repo,
         unless_disabled(T.Feeds.SeenPruner),
+        unless_disabled(T.Matches.MatchesTimeCheck),
```
ruslandoga, line 29:
```suggestion
        unless_disabled(T.Matches.MatchExpirer),
```

lib/t/calls.ex
```diff
@@ -106,6 +106,19 @@ defmodule T.Calls do
...
     Bot.async_post_message(m)
 
+    match_id =
+      Match
+      |> where([m], m.user_id_1 == ^caller and m.user_id_2 == ^called)
+      |> or_where([m], m.user_id_1 == ^called and m.user_id_2 == ^caller)
+      |> order_by(desc: :inserted_at)
+      |> limit(1)
+      |> select([m], m.id)
+      |> Repo.one()
```
ruslandoga, line 116:
```suggestion
      |> Repo.one!()
```

lib/t/calls.ex
```diff
@@ -106,6 +106,19 @@ defmodule T.Calls do
     Logger.warn(m)
     Bot.async_post_message(m)
 
+    match_id =
```
ruslandoga, line 109: Ideally it would be loaded (and even more ideally inserted into events) in `{caller, called}` query above.

lib/t/feeds.ex
```diff
@@ -226,6 +234,7 @@ defmodule T.Feeds do
     |> not_liked_profiles_q(user_id)
     |> not_liker_profiles_q(user_id)
     |> not_seen_profiles_q(user_id)
+    |> not_expired_match_profiles_q(user_id)
```
ruslandoga, line 237:
```suggestion
    |> not_expired_match_with_q(user_id)
```
or something similar `not_expired_match_profiles_q` doesn't sound right

lib/t/matches.ex
```diff
@@ -20,6 +20,8 @@ defmodule T.Matches do
   @pubsub T.PubSub
   @topic "__m"
 
+  @match_expiration_duration 172_800
```
ruslandoga, line 23: What's 172_800? Also values like this shouldn't be hardcoded but passed as args.

lib/t/matches.ex
```diff
@@ -125,7 +138,16 @@ defmodule T.Matches do
...
         m = "New match: #{user_id_1} and #{user_id_2}"
         Bot.async_post_message(m)
-        Repo.insert(%Match{user_id_1: user_id_1, user_id_2: user_id_2})
+        {:ok, match} = Repo.insert(%Match{user_id_1: user_id_1, user_id_2: user_id_2})
+
+        Repo.insert(%MatchEvent{
+          timestamp: DateTime.utc_now() |> DateTime.truncate(:second),
+          match_id: match.id,
+          event: "created"
+        })
```
ruslandoga, line 147: Both inserts should be performed in a single transaction.

lib/t/matches.ex
```diff
@@ -125,7 +138,16 @@ defmodule T.Matches do
...
+
+        expiration_date = expiration_date(match.id)
+        {:ok, %Match{match | expiration_date: expiration_date}}
```
ruslandoga, line 150: `expiration_date` appears to be a view concern, try moving it to `TWeb.MatchView`

lib/t/matches.ex
```diff
@@ -180,6 +202,16 @@ defmodule T.Matches do
...
+    |> with_expiration_date()
+  end
+
+  @spec list_expired_matches(uuid) :: [%Match{}]
+  def list_expired_matches(user_id) do
+    ExpiredMatch
+    |> where([m], m.user_id == ^user_id)
+    |> order_by(asc: :inserted_at)
+    |> Repo.all()
+    |> preload_expired_match_profiles()
```
ruslandoga, line 214: Better to use join here than preload.

lib/t/matches.ex
```diff
@@ -307,6 +376,41 @@ defmodule T.Matches do
     end)
   end
 
+  defp with_expiration_date(matches) do
```
ruslandoga, line 379: This function lists all events for all matches, one by one, that's wasteful. Please check out lateral jons, it can be done in a single query when listing matches.

lib/t/matches.ex
```diff
@@ -325,12 +429,41 @@ defmodule T.Matches do
     end)
   end
 
+  defp insert_expired_match(multi) do
+    Multi.run(multi, :insert_expired_match, fn _repo, %{unmatch: unmatch} ->
+      %{id: match_id, users: [user_id_1, user_id_2]} = unmatch
```
ruslandoga, line 434: Not quite clear why unmatching results in an expired match.

lib/t/matches.ex
```diff
@@ -357,6 +490,8 @@ defmodule T.Matches do
     m = "Saving slots offer for match: #{offerer_id} offered date  #{mate_id}"
     Bot.async_post_message(m)
 
+    match_timeslot_new_event(match_id, "saving slot")
```
ruslandoga, line 493: Best if event names don't have whitespace.

ruslandoga: `new` in function names usually means building a data structure in memory, not persisting to DB. If it's persisting to DB it should be renamed and moved to the same transaction as other ops below.

lib/t/matches.ex
```diff
@@ -511,7 +659,7 @@ defmodule T.Matches do
     :ok
   end
 
-  defp get_match_id_for_users!(user_id_1, user_id_2) do
+  def get_match_id_for_users!(user_id_1, user_id_2) do
```
ruslandoga, line 662: Why does it need to be public?

lib/t/matches.ex
```diff
@@ -593,4 +741,96 @@ defmodule T.Matches do
...
+
+    for uid <- user_ids do
+      broadcast_for_user(uid, message)
+    end
+  end
+
+  # Expired Matches
+
+  def match_timeslot_new_event(match_id, event) do
+    if match_id != [] and match_id != nil and match_id != "" do
```
ruslandoga, line 756: Why `match_id` can be so many types? Should be valid uuid. There is no reason it can't be known at call site.

lib/t/matches.ex
```diff
@@ -593,4 +741,96 @@ defmodule T.Matches do
...
+  end
+
+  def match_expired_check() do
+    expired_matches_q()
+    |> join(:inner, [m], ma in Match, on: ma.id == m.match_id)
+    |> where([m], m.timestamp < fragment("now() - INTERVAL '48 hours'"))
+    |> select([m, ma], {ma.user_id_1, ma.user_id_2})
+    |> T.Repo.all()
+    |> Enum.map(fn {user_id_1, user_id_2} ->
+      expire_match(user_id_1, user_id_2)
```
ruslandoga, line 772: Please do this a single query, it's wasteful otherwise.

lib/t/matches.ex
```diff
@@ -593,4 +741,96 @@ defmodule T.Matches do
...
+    Oban.insert(job)
+  end
+
+  def mark_expired_match_seen(match_id, by_user_id) do
+    ExpiredMatch
+    |> where(match_id: ^match_id, user_id: ^by_user_id)
+    |> Repo.delete_all()
+  end
+
+  def expiration_date(match_id) do
```
ruslandoga, line 799: Please remove this function and use a lateral join.

lib/t/matches/match_event.ex
```diff
@@ -0,0 +1,12 @@
+defmodule T.Matches.MatchEvent do
+  @moduledoc false
+  use Ecto.Schema
+
+  @primary_key false
+  @foreign_key_type Ecto.Bigflake.UUID
```
ruslandoga, line 6: Foreigns keys are not used here.

lib/t/matches/matches_time_check.ex
```diff
@@ -0,0 +1,28 @@
...
+  """
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @impl true
+  def init(opts) do
+    check_interval = opts[:check_interval] || :timer.minutes(1)
+    :timer.send_interval(check_interval, :prune)
+    {:ok, %{check_interval: check_interval}}
```
ruslandoga, line 15: `check_interval` is not used after `init`, doesn't have to be in the state.

lib/t/push_notifications/apns.ex
```diff
@@ -77,6 +77,15 @@ defmodule T.PushNotifications.APNS do
     base_alert_payload(type, alert)
   end
 
+  def build_alert_payload("match_about_to_expire" = type, _data) do
+    alert = %{
+      "title" => dgettext("apns", "Your match is about to expire 😢"),
+      "body" => dgettext("apns", "Invite your match to date if you want to keep it alive ✨")
```
ruslandoga, line 83: maybe `to a date` or `for a call`

lib/t_web/channels/feed_channel.ex
```diff
@@ -6,6 +6,8 @@ defmodule TWeb.FeedChannel do
   alias T.Feeds.{FeedProfile}
   alias T.{Feeds, Calls, Matches, Accounts}
 
+  @match_expiration_duration 172_800
```
ruslandoga, line 9: I think it's already been declared somewhere.

lib/t_web/channels/feed_channel.ex
```diff
@@ -32,11 +34,18 @@ defmodule TWeb.FeedChannel do
...
         |> render_matches(screen_width)
 
+      expired_matches =
+        user_id
+        |> Matches.list_expired_matches()
+        |> render_expired_matches(screen_width)
+
       reply =
         %{}
+        |> Map.put("match_expiration_duration", @match_expiration_duration)
```
ruslandoga, line 44: I'd call it [`match_ttl`](https://en.wikipedia.org/wiki/Time_to_live), at least that's how resources with limited life-span are usually called.

lib/t_web/channels/feed_channel.ex
```diff
@@ -74,6 +83,12 @@ defmodule TWeb.FeedChannel do
     {:reply, :ok, socket}
   end
 
+  def handle_in("seen", %{"expired_match_id" => match_id}, socket) do
+    by_user_id = me_id(socket)
+    Matches.mark_expired_match_seen(match_id, by_user_id)
```
ruslandoga, line 88: Doesn't it delete it? Then naming is misleading.

priv/repo/migrations/20211103061515_add_expired_matches_table.exs
```diff
@@ -0,0 +1,17 @@
...
+  def change do
+    create table(:expired_matches, primary_key: false) do
+      add :match_id, :uuid, null: false
+      add :user_id, references(:users, @opts), null: false
+      add :with_user_id, references(:users, @opts), null: false
+
+      timestamps(updated_at: false)
+    end
+
+    create unique_index(:expired_matches, [:match_id, :user_id])
```
ruslandoga, line 15: Why not `[user_id, match_id]`? I think the table is queried by user_id to get matches, not the other way around.

test/t/accounts/blocking_test.exs
```diff
@@ -4,6 +4,8 @@ defmodule T.Accounts.BlockingTest do
   alias T.Accounts.UserReport
   alias T.Matches.Match
 
+  @match_expiration_duration 172_800
```
ruslandoga, line 7: hardcoding again.

ruslandoga (changes requested):
Note that I didn't read through all of the changes since I guess it's not finished yet.

My biggest concern right now is the current approach for querying for expired matches results in a `n + 1` queries. An approach like a lateral join of matches on match_events with filtering for the last event of certain kind would be more efficient both in DB processing and queries count.

Other comments are mostly in regards to clarity, which is still important but not as much as a good general approach.

lib/t/application.ex
ruslandoga: It's not clear what `MatchesTimeCheck` does judging just by its name.

lib/t/matches.ex
ruslandoga: Maybe `Multi.insert` or `Multi.insert_all` functions could be used here?

ruslandoga (approved):

lib/t/calls.ex
```diff
@@ -93,23 +94,54 @@ defmodule T.Calls do
...
+        m =
+          "New call starts: #{fetch_name(caller)}(#{caller}) and #{fetch_name(called)}(#{called})"
 
-    m = "New call starts: #{fetch_name(caller)}(#{caller}) and #{fetch_name(called)}(#{called})"
+        Logger.warn(m)
+        Bot.async_post_message(m)
 
-    Logger.warn(m)
-    Bot.async_post_message(m)
+        T.Matches.insert_match_event(match_id, "call_start")
```
akhrail1996, line 138: can be moved to Multi.above

lib/t/matches.ex
```diff
@@ -125,7 +128,22 @@ defmodule T.Matches do
         [user_id_1, user_id_2] = Enum.sort(user_ids)
         m = "New match: #{user_id_1} and #{user_id_2}"
         Bot.async_post_message(m)
-        Repo.insert(%Match{user_id_1: user_id_1, user_id_2: user_id_2})
+
+        Multi.new()
+        |> Multi.insert(:match, %Match{user_id_1: user_id_1, user_id_2: user_id_2})
+        |> Multi.insert(:match_event, fn %{match: match} ->
+          %MatchEvent{
```
akhrail1996, line 135: not sure if Multi inside Multi is good here, maybe they could be united via some .insert_all()

lib/t/matches.ex
```diff
@@ -182,6 +200,19 @@ defmodule T.Matches do
...
+  @spec list_expired_matches(uuid) :: [%Match{}]
+  def list_expired_matches(user_id) do
+    ExpiredMatch
+    |> where([m], m.user_id == ^user_id)
+    |> order_by(asc: :inserted_at)
+    |> join(:inner, [m], p in FeedProfile, on: m.with_user_id == p.user_id)
+    |> select([m, p], {m, p})
+    |> Repo.all()
+    |> Enum.map(fn {match, feed_profile} ->
+      %ExpiredMatch{match | profile: feed_profile}
```
akhrail1996, line 212: gotta check that profiles are rendered correctly

lib/t/matches.ex
```diff
@@ -272,6 +303,43 @@ defmodule T.Matches do
...
+  end
+
+  defp notify_expired(users, match_id) do
+    for user_id <- users do
+      broadcast_for_user(user_id, {__MODULE__, :expired, match_id})
+    end
+  end
+
   # TODO cleanup
   defp preload_match_profiles(matches, user_id) do
```
akhrail1996, line 344: probably could be :joined and not preloaded similar to ExpiredMatch

test/t/matches/timeslots_test.exs
```diff
@@ -212,6 +212,19 @@ defmodule T.Matches.TimeslotsTest do
 
       assert length(match_event) == 2
     end
+
+    test "decline", %{profiles: [_p1, p2], match: match} do
```
akhrail1996, line 216: not clear what exactly is tested here from the name

config/runtime.exs
```diff
@@ -136,13 +136,6 @@ if config_env() == :dev do
         team_id: System.fetch_env!("APNS_TEAM_ID"),
         topic: System.fetch_env!("APNS_TOPIC"),
         env: :dev
-      },
-      %{
-        key: System.fetch_env!("PROD_APNS_KEY"),
-        key_id: System.fetch_env!("PROD_APNS_KEY_ID"),
-        team_id: System.fetch_env!("APNS_TEAM_ID"),
-        topic: System.fetch_env!("APNS_TOPIC"),
-        env: :prod
```
akhrail1996, line 145: deleted to not accidentally send pushes to prod while in dev
@ruslandoga, or are pushes to prod disabled in some other place as well?

ruslandoga: no, you are right, prod pushes can be sent from dev env

lib/t/matches.ex
akhrail1996: not that easy since you gotta know which mate is that (uuid1 or uuid2) - maybe later if need

test/t/matches/timeslots_test.exs
akhrail1996: doesn't really test much + does not belong here (in case of tests above we added MatchEvent check as extra one, but here nothing besides MatchEvent insertion is tested) => if inserted it would probably go to some separate MatchEventTest.exs

test/t/matches/expired_match_test.exs
```diff
@@ -0,0 +1,123 @@
...
+               %Oban.Job{
+                 args: %{
+                   "match_id" => ^match_id,
+                   "type" => "match_about_to_expire"
+                 }
+               }
+             ] = all_enqueued(worker: T.PushNotifications.DispatchJob)
+    end
+
+    test "call is not the latest event" do
```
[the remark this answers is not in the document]
akhrail1996: also doesn't add much to "match with call is not expired" test above

akhrail1996:
@ruslandoga re-requesting review for you to maybe find something bad before merging to master

lib/t/calls.ex
```diff
@@ -93,23 +94,64 @@ defmodule T.Calls do
...
+    Multi.insert(multi, :event, fn %{match_id: match_id} ->
+      %T.Matches.MatchEvent{
+        timestamp: DateTime.truncate(DateTime.utc_now(), :second),
+        match_id: match_id,
+        event: "call_start"
+      }
+    end)
+  end
+
   @spec accept_call(Ecto.UUID.t(), DateTime.t()) :: :ok
```
ruslandoga, line 134: It no longer returns `:ok`.

ruslandoga:
Screenshot 2021-11-09 at 13 49 41

This needs to be fixed. Locally it makes my tests fail.

ruslandoga:
Warnings need to be fixed as well:
Screenshot 2021-11-09 at 13 52 31

lib/t/calls.ex
```diff
@@ -93,23 +94,64 @@ defmodule T.Calls do
...
 
-    {1, _} =
-      Call
-      |> where(id: ^call_id)
-      |> Repo.update_all(set: [accepted_at: now])
+      {:error, _step, _reason, _changes} = failure ->
+        failure
+    end
 
     :ok
```
ruslandoga, line 156: Ah, it returns `:ok` here. That's not good since you are ignoring possible error results from the transaction above.

ruslandoga:
All in all, It seems fine to me. Good job 👍

akhrail1996:
> This needs to be fixed, otherwise it'd lead to falky tests. Locally it already makes my tests fail, I don't know the exact reason but it seems that the error changes something for postgres which makes it change the order of rows in some queries that don't provide explicit `order_by`

fixed the test fail, but not dbConnectionError - still occurs every time:
image
Not sure how to debug it (if its needed at all), any ideas?

akhrail1996:
@ruslandoga thanks for the review and your kind words :)

test/t/matches/expired_match_test.exs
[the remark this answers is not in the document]
orlovvlad95: Ok, will fix the name
In my opinion there is a different with test above, it tests cases, when call_start is not the latest event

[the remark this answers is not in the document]
akhrail1996: I deleted the test, so no need to change its name :)
The check for call_start doesn't depend on the order, so this is duplicating "match with call is not expired" test above

PR #388, opened 2021-11-03
expire matches [rails part]

akhrail1996:


lib/t/calls.ex
```diff
@@ -117,6 +117,8 @@ defmodule T.Calls do
 
     T.Matches.match_timeslot_new_event(match_id, "call start")
 
+    # TODO BROADCAST RESET EXPIRATION DATE
```
akhrail1996, line 120: will do this myself bit later

lib/t/matches.ex
```diff
@@ -601,24 +736,75 @@ defmodule T.Matches do
...
+    |> where(match_id: ^match_id)
+    |> distinct([e], e.match_id)
+    |> order_by(desc: :timestamp)
+    |> select([e], e.timestamp)
+    |> Repo.one()
+    # TODO TO ENV VARS
+    |> DateTime.add(2 * 24 * 60 * 60)
+  end
+
+  # defp schedule_match_about_to_expire(match_id) do
```
akhrail1996, line 806: (commented it out to silence warning about function being unused)

@orlovvlad95 could you please create a pruner system (similar to the one which expires matches) which will send push notifications using this function 2 hours before match is about to expire?

akhrail1996:
@orlovvlad95 if you have time after push-notification scheduling system, you can try to add some tests (would be probably best for training if you would come up with "what to test" yourself as well 😀)

lib/t/matches.ex
```diff
@@ -125,7 +137,16 @@ defmodule T.Matches do
...
-        Repo.insert(%Match{user_id_1: user_id_1, user_id_2: user_id_2})
+        {:ok, match} = Repo.insert(%Match{user_id_1: user_id_1, user_id_2: user_id_2})
+
+        Repo.insert(%MatchEvent{
+          timestamp: DateTime.utc_now() |> DateTime.truncate(:second),
+          match_id: match.id,
+          event: "created"
+        })
+
+        expiration_date = expiration_date(match.id)
```
akhrail1996, line 148: alternatively we can just compute it here using DateTime.now + expiration_duration for better performance (but it may hurt consistency a bit)

lib/t/calls.ex
```diff
@@ -117,6 +117,8 @@ defmodule T.Calls do
 
     T.Matches.match_timeslot_new_event(match_id, "call start")
 
+    T.Matches.notify_match_expiration_reset(match_id, [caller, called])
```
akhrail1996, line 120: needed for client to remove expiration_timer in UI

lib/t/feeds.ex
```diff
@@ -226,6 +234,7 @@ defmodule T.Feeds do
     |> not_liked_profiles_q(user_id)
     |> not_liker_profiles_q(user_id)
     |> not_seen_profiles_q(user_id)
+    |> not_expired_match_profiles_q(user_id)
```
akhrail1996, line 237: not returning expired matches to initial feed (only after users "seens" them)

lib/t/matches.ex
```diff
@@ -66,14 +66,26 @@ defmodule T.Matches do
...
-    broadcast_from_for_user(by_user_id, {__MODULE__, :matched, %{id: match_id, mate: user_id}})
-    broadcast_for_user(user_id, {__MODULE__, :matched, %{id: match_id, mate: by_user_id}})
+  defp maybe_notify_match(
+         %Match{id: match_id, expiration_date: expiration_date},
+         by_user_id,
+         user_id
+       ) do
+    broadcast_from_for_user(
+      by_user_id,
+      {__MODULE__, :matched, %{id: match_id, mate: user_id, expiration_date: expiration_date}}
```
akhrail1996, line 78: adding expiration_date to every push / reply for client to always have the updated expiration_date

lib/t/matches.ex
```diff
@@ -180,6 +202,16 @@ defmodule T.Matches do
     |> Repo.all()
     |> preload_match_profiles(user_id)
     |> with_timeslots(user_id)
+    |> with_expiration_date()
+  end
+
+  @spec list_expired_matches(uuid) :: [%Match{}]
```
akhrail1996, line 208: returning expiredMatches as a separate category in feedChannel.join (likes + matches + missed_calls + expired_matches)

lib/t/matches.ex
```diff
@@ -272,6 +304,43 @@ defmodule T.Matches do
...
+      |> where(user_id_2: ^user_id_2)
+      |> select([m], %{id: m.id, users: [m.user_id_1, m.user_id_2]})
+      |> Repo.delete_all()
+      |> case do
+        {1, [match]} -> {:ok, match}
+        {0, _} -> {:error, :match_not_found}
+      end
+    end)
+    |> delete_likes()
+    |> insert_expired_match()
```
akhrail1996, line 326: need to insert this match to ExpiredMatches table

lib/t/matches.ex
```diff
@@ -307,6 +376,41 @@ defmodule T.Matches do
     end)
   end
 
+  defp with_expiration_date(matches) do
```
akhrail1996, line 379: adding expiration_date for every match

lib/t/matches.ex
```diff
@@ -325,12 +429,41 @@ defmodule T.Matches do
...
   end
 
+  defp insert_expired_match(multi) do
+    Multi.run(multi, :insert_expired_match, fn _repo, %{unmatch: unmatch} ->
+      %{id: match_id, users: [user_id_1, user_id_2]} = unmatch
+
+      timestamp = DateTime.utc_now() |> DateTime.truncate(:second) |> DateTime.to_naive()
+
+      {count, _} =
+        Repo.insert_all(
```
akhrail1996, line 439: inserting two ExpiredMatch entries mirroring each other to then delete each of them on "seen" by respective user

lib/t/matches.ex
```diff
@@ -600,25 +742,95 @@ defmodule T.Matches do
...
+    |> Enum.map(fn {user_id_1, user_id_2} ->
+      expire_match(user_id_1, user_id_2)
+    end)
+  end
+
+  def match_soon_to_expire_check() do
+    expired_matches_q()
+    |> join(:inner, [m], ma in Match, on: ma.id == m.match_id)
+    |> where([m], m.timestamp < fragment("now() - INTERVAL '46:1 HOURS:MINUTES'"))
+    |> where([m], m.timestamp > fragment("now() - INTERVAL '46 hours'"))
```
akhrail1996, line 780: not sure whether such approach will work reliably given the pruner interval is also 1 minute

akhrail1996: the problem is that we need to send this push only once (so e.g. expanding to 46:2 may result in two pushes)

lib/t_web/channels/feed_channel.ex
```diff
@@ -32,11 +34,18 @@ defmodule TWeb.FeedChannel do
...
         |> render_matches(screen_width)
 
+      expired_matches =
+        user_id
+        |> Matches.list_expired_matches()
+        |> render_expired_matches(screen_width)
+
       reply =
         %{}
+        |> Map.put("match_expiration_duration", @match_expiration_duration)
```
akhrail1996, line 44: needed for client to always have an up-to-date match_expiration_duration which enable to correctly show % of time left till expiration (i.e. 10 / 48 hours = circle with 21% filled)

test/t/accounts/blocking_test.exs
```diff
@@ -27,9 +29,11 @@ defmodule T.Accounts.BlockingTest do
       assert {:ok, %{match: nil}} = Matches.like_user(reporter.id, reported.id)
       assert {:ok, %{match: %Match{id: match_id}}} = Matches.like_user(reported.id, reporter.id)
 
+      exp_date = expiration_date()
```
akhrail1996, line 32: needed here and everywhere below to record the time wen the push / event occurred (since in push_receive assertion phase the timestamp sometimes is already 1 second up)

even this approach doesn't look ideal, however never managed to break it yet

PR #389, opened 2021-11-03
update erlang to 24.1.4

ruslandoga:
example how to update erlang/elixir/alpine version, note that it's updated in two places, ci (to run tests on the relevant version combo) and in dockerfile (to actually update the version combo)

if alpine version is updated, it'd also be updated [in the final image](https://github.com/getsince/test3/blob/93f91d996961321712b7eaa761768728e2576021/Dockerfile#L39)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,7 +28,7 @@ jobs:
         uses: erlef/setup-elixir@v1
         with:
           elixir-version: "1.12.3"
-          otp-version: "24.1.2"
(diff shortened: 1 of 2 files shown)


akhrail1996 (approved):
thanks!
does dependabot alert on these updates? 
do you recommend to update every time?

ruslandoga:
> does dependabot alert on these updates?

no, but it can be setup

> do you recommend to update every time?

it depends on the particular release, bugfix releases should probably be updated asap

PR #400, opened 2021-11-09
add location & age filters to feed

akhrail1996:
- [x] extract birthdate from profile.story in prod db
- [ ] [upon release] send push asking to add "age" sticker to story for those with birthdate == nil
- [x] add tests


akhrail1996:
@ruslandoga could you please quickly scan through the changes and tell if you have any general comments on the chosen approach so far?

lib/t/accounts.ex
```diff
@@ -547,6 +551,28 @@ defmodule T.Accounts do
...
+      %Profile{filters: %Profile.Filters{distance: new_distance}} = new_profile
+
+      Ecto.Changeset.change(%DistancePreference{user_id: user_id, distance: new_distance},
+        distance: new_distance
+      )
+    end)
+  end
+
+  defp update_profile_age_preferences(multi, %Profile{user_id: user_id}) do
+    Multi.insert_or_update(multi, :update_profile_age_preferences, fn %{profile: new_profile} ->
```
ruslandoga, line 565: Should probably be an insert with `on_conflict: :replace_all`

lib/t/accounts/age_preference.ex
```diff
@@ -0,0 +1,11 @@
+defmodule T.Accounts.AgePreference do
+  use Ecto.Schema
+
+  @primary_key false
+  @foreign_key_type Ecto.Bigflake.UUID
+  schema "age_preferences" do
+    field(:user_id, Ecto.UUID, primary_key: true)
```
ruslandoga, line 7: Something's wrong with your formatter setup, `field` macros don't require brackets. I'ts not a problem here, but

lib/t/accounts/profile.ex
```diff
@@ -5,27 +5,30 @@ defmodule T.Accounts.Profile do
   @primary_key false
   @foreign_key_type Ecto.Bigflake.UUID
   schema "profiles" do
-    belongs_to :user, T.Accounts.User, primary_key: true
+    belongs_to(:user, T.Accounts.User, primary_key: true)
 
-    field :story, {:array, :map}
-    field :location, Geo.PostGIS.Geometry
+    field(:story, {:array, :map})
+    field(:location, Geo.PostGIS.Geometry)
```
ruslandoga, line 11: ...here it results in unnecessary diffs. Many other files would be changed as well.

lib/t/accounts/profile.ex
```diff
@@ -5,27 +5,30 @@ defmodule T.Accounts.Profile do
...
 
     # TODO remove
     embeds_one :filters, Filters, primary_key: false, on_replace: :delete do
       # ["F"] or ["F", "M"], etc.
-      field :genders, {:array, :string}
+      field(:genders, {:array, :string})
```
ruslandoga, line 16: These need to be casted in `essential_info_changeset` in `cast_embed(:filters,`, although I'd remove the embed, it was a quick workaround when adding genders, it's not really used anymore.

lib/t/feeds.ex
```diff
@@ -51,8 +51,8 @@ defmodule T.Feeds do
   end
 
   # TODO accept cursor
-  @spec list_received_likes(Ecto.UUID.t(), Geo.Point.t()) :: [feed_profile]
-  def list_received_likes(user_id, location) do
+  @spec list_received_likes(Ecto.UUID.t()) :: [feed_profile]
```
ruslandoga, line 54: `[feed_profile]` would probably become `[{feed_profile, non_neg_integer}]`

lib/t/feeds.ex
```diff
@@ -85,23 +85,46 @@ defmodule T.Feeds do
   ### Feed
 
   @type feed_cursor :: DateTime.t()
-  @type feed_profile :: {%FeedProfile{}, distance_km :: non_neg_integer}
+  @type feed_profile :: %FeedProfile{}
```
ruslandoga, line 88: This is confusing, feed profile is feed profile. I try to pick different names for composite types and not name the composite type by one of its elements.

lib/t/feeds.ex
```diff
@@ -85,23 +85,46 @@ defmodule T.Feeds do
...
           {[feed_profile], feed_cursor}
-  def fetch_feed(user_id, location, gender, gender_preferences, count, feed_cursor) do
+  def fetch_feed(
+        user_id,
+        location,
+        gender,
+        gender_preferences,
+        min_age,
+        max_age,
+        distance,
```
ruslandoga, line 109: too many arguments, easy to make a typo, try using some collection like a list or a map or maybe even a struct to put them in, `def fetch_feed(user_id, location, gender, gender_preference: gender_preference, min_age: min_age, max_age: max_age, distance: distance)`

ruslandoga:
```elixir
@spec fetch_feed(Ecto.UUID.t(), Geo.Point.t(), String.t, Keyword.t, ...etc) :: {[feed_profile], feed_cursor}
def fetch_feed(user_id, location, gender, filters, count, feed_cursor) do
  gender_preference = filters[:gender_preference] || raise "missing gender preference filter"
  min_age = filters[:min_age] || raise "missing min age filter"
  # etc.
  continue_feed(..., filters, ...)
end 
```

lib/t/feeds.ex
```diff
@@ -115,42 +138,97 @@ defmodule T.Feeds do
...
+      query
+    end
+  end
+
+  defp maybe_apply_max_age_filer(query, max_age) do
+    if max_age do
+      %{year: y, month: m, day: d} = DateTime.utc_now()
+      oldest = %Date{year: y - String.to_integer(max_age), month: m, day: d}
+
+      where(query, [p], p.birthdate >= ^oldest)
```
ruslandoga, line 211: Should it be reciprocal, that is to check if other user is interested in current user?

lib/t_web/channels/feed_channel.ex
```diff
@@ -222,15 +224,13 @@ defmodule TWeb.FeedChannel do
...
-    assigns = [profile: profile, screen_width: screen_width, distance: distance]
-    render(FeedView, "feed_item.json", assigns)
+  defp render_feed_profile(profile, screen_width) do
+    assigns = [profile: profile, screen_width: screen_width]
+    render(FeedView, "feed_profile.json", assigns)
   end
 
   defp render_feed(feed, screen_width) do
-    Enum.map(feed, fn feed_item -> render_feed_item(feed_item, screen_width) end)
+    Enum.map(feed, fn feed_profile -> render_feed_profile(feed_profile, screen_width) end)
```
ruslandoga, line 233: I'd keep it as feed item, since feed profile doesn't really have a distance (that user didn't submit it).

priv/repo/migrations/20211109083906_add_age_and_distance_preferences.exs
```diff
@@ -0,0 +1,18 @@
+defmodule T.Repo.Migrations.AddAgeAndDistancePreferences do
+  use Ecto.Migration
+
+  @opts [on_delete: :delete_all, type: :uuid]
+
+  def up do
+    create table(:age_preferences, primary_key: false) do
+      add(:user_id, references(:users, @opts), primary_key: true)
+      add(:min_age, :integer, null: true, default: nil)
```
ruslandoga, line 9: `null: true` is a bad idea.

priv/repo/migrations/20211109083906_add_age_and_distance_preferences.exs
```diff
@@ -0,0 +1,18 @@
...
+      add(:max_age, :integer, null: true, default: nil)
+    end
+
+    create table(:distance_preferences, primary_key: false) do
+      add(:user_id, references(:users, @opts), primary_key: true)
+      add(:distance, :integer, null: true, default: nil)
```
ruslandoga, line 15: If it's null, just don't store it.

lib/t/feeds.ex
ruslandoga: Since I later saw that filters are optional, there is no need to `raise` in case they are missing and `filters` can be passed unchecked to `continue_feed` where filters can be passed to each `maybe_filter` function and apply where clauses only if `if max_age = filters[:max_age] do`.

lib/t/accounts/age_preference.ex
akhrail1996: yeah, very weird
I even tried setting default formatter, but it didn't help:
image

akhrail1996: my local formatter.exs seem to be the same as here in master:
image

maybe some deps bump led to this? 🤔

lib/t/accounts/profile.ex
akhrail1996: @ruslandoga so maybe use :virtual instead?

akhrail1996: or maybe just remove `filters` column from `profiles` table altogether?

ruslandoga: 2

lib/t/feeds.ex
akhrail1996: nope (at least for now) - will narrow selection much

lib/t_web/channels/feed_channel.ex
akhrail1996: actually `feed_profile` is really a feed_profile since I deleted distance from `feed_item`

lib/t/feeds.ex
```diff
@@ -115,42 +138,97 @@ defmodule T.Feeds do
...
+    else
+      query
+    end
+  end
+
+  defp maybe_apply_distance_filter(query, location, distance) do
+    if distance do
+      where(query, [p], distance_km(^location, p.location) <= ^distance)
```
ruslandoga, line 219: btw this is too slow since it can't use any index

lib/t_web/channels/feed_channel.ex
ruslandoga: ah, so there is only filtering now and the distance itself is not sent to the user?

lib/t/feeds.ex
ruslandoga: https://github.com/bryanjos/geo_postgis/blob/80ebb06b03578ec0e170de31ce734674ca79cb69/lib/geo_postgis.ex#L62-L70

ruslandoga: so `st_dwithin_in_meters` seems to be fine

ruslandoga: more info https://postgis.net/docs/ST_DWithin.html

akhrail1996: any easy fix in your mind?

akhrail1996: @ruslandoga  to clarify - current code is fine in the end of the day because it already uses `st_dwithin_in_meters`?

lib/t_web/channels/feed_channel.ex
akhrail1996: yes

lib/t/feeds.ex
ruslandoga: no?

lib/t/accounts.ex
```diff
@@ -425,7 +430,31 @@ defmodule T.Accounts do
...
+        [] -> nil
+        some -> some
+      end
+
+    %Profile{
+      profile
+      | gender_preference: gender_preference,
+        min_age: min_age,
+        max_age: max_age,
+        distance: distance
```
akhrail1996, line 456: need this to conveniently render OwnProfile afterwards

priv/repo/migrations/20211109083906_add_age_and_distance_preferences.exs
```diff
@@ -0,0 +1,43 @@
...
+
+    T.Repo.insert_all(
+      "max_age_preferences",
+      user_ids |> Enum.map(fn id -> %{user_id: id, age: 100} end)
+    )
+
+    T.Repo.insert_all(
+      "distance_preferences",
+      user_ids |> Enum.map(fn id -> %{user_id: id, distance: 20000} end)
+    )
```
akhrail1996, line 41: it's easier (& maybe even simpler?) when all users have maxed out filters -> no need to account for nil values (which do not bring any additional value)

test/support/factory.ex
```diff
@@ -107,7 +107,10 @@ defmodule T.Factory do
       gender: gender,
       name: opts[:name] || "that",
       times_liked: opts[:times_liked] || 0,
-      filters: %{genders: opts[:accept_genders] || ["F"]}
+      gender_preference: opts[:accept_genders] || ["F"],
+      distance: opts[:distance] || 20000,
+      max_age: opts[:max_age] || 100,
+      min_age: opts[:min_age] || 18
```
akhrail1996, line 113: since these are now non-nil

test/t_web/channels/feed_channel_test.exs
```diff
@@ -311,22 +307,19 @@ defmodule TWeb.FeedChannelTest do
...
+                 story: [
+                   %{
+                     "background" => %{
+                       "proxy" =>
+                         "https://d1234.cloudfront.net/1hPLj5rf4QOwpxjzZB_S-X9SsrQMj0cayJcOCmnvXz4/fit/1000/0/sm/0/aHR0cHM6Ly9wcmV0ZW5kLXRoaXMtaXMtcmVhbC5zMy5hbWF6b25hd3MuY29tL3Rlc3Q",
+                       "s3_key" => "test"
+                     },
+                     "labels" => []
+                   }
+                 ]
```
akhrail1996, line 322: this change {distance: ..., profile: ...}  -> profile {...} does not affect frontend: https://github.com/getsince/test5/blob/rails-fork/Since/Feed/Models/FeedProfile.swift#L41

priv/repo/migrations/20211109083906_add_age_and_distance_preferences.exs
ruslandoga: not really

test/t_web/channels/feed_channel_test.exs
ruslandoga: Do you mean `{distance: 1000, profile: {}}` vs `{profile: {}}`?

lib/t/accounts.ex
ruslandoga: you can do this in `select`

ruslandoga:
Seems like there is one type of filter per profile, since you are not doing reciprocal filtering, you can put filters in profile table and add indexes there.

lib/t/accounts.ex
```diff
@@ -425,7 +430,31 @@ defmodule T.Accounts do
...
   def get_profile!(user_id) when is_binary(user_id) do
-    Repo.get!(Profile, user_id)
+    {profile, min_age, max_age, distance} =
+      Profile
+      |> where([p], p.user_id == ^user_id)
+      |> join(:left, [p], min in MinAgePreference, on: p.user_id == min.user_id)
+      |> join(:left, [p, min], max in MaxAgePreference, on: p.user_id == max.user_id)
+      |> join(:left, [p, min, max], d in DistancePreference, on: p.user_id == d.user_id)
+      |> select([p, min, max, d], {p, min.age, max.age, d.distance})
+      |> Repo.one()
```
ruslandoga, line 440:
```
|> Repo.one!()
```

ruslandoga (changes requested):
Too complicated, try colocating age and distance filters with profile info.

Create index on filters and location.

test/t_web/channels/feed_channel_test.exs
akhrail1996: nope, `{distance: 1000, profile: {}}` vs `{}`

ruslandoga: then frontend is affected

lib/t/accounts.ex
```diff
@@ -425,7 +427,21 @@ defmodule T.Accounts do
...
+      Profile
+      |> where([p], p.user_id == ^user_id)
+      |> Repo.one!()
+
+    gender_preference =
+      case GenderPreference
+           |> where([g], g.user_id == ^user_id)
+           |> select([g], g.gender)
+           |> Repo.all() do
+        [] -> nil
```
ruslandoga, line 440: no need to replace empty list with `nil`

lib/t/accounts.ex
```diff
@@ -442,7 +458,7 @@ defmodule T.Accounts do
     |> Multi.update(:profile, fn %{user: %{profile: profile}} ->
       Profile.changeset(profile, attrs, validate_required?: true)
     end)
-    |> update_profile_gender_preferences(profile)
+    |> update_profile_gender_preference(profile, attrs)
```
ruslandoga, line 461: it's still plural

lib/t/accounts.ex
```diff
@@ -458,11 +474,22 @@ defmodule T.Accounts do
     end)
     |> Repo.transaction()
     |> case do
-      {:ok, %{user: user, profile: %Profile{} = profile}} ->
+      {:ok,
+       %{
+         user: user,
+         profile: %Profile{} = profile,
+         update_profile_gender_preference: genders
```
ruslandoga, line 481: rename multi to `gender_preferences`

lib/t/accounts.ex
```diff
@@ -488,48 +515,71 @@ defmodule T.Accounts do
     end)
     |> Repo.transaction()
     |> case do
-      {:ok, %{profile: profile}} -> {:ok, profile}
-      {:error, :profile, %Ecto.Changeset{} = changeset, _changes} -> {:error, changeset}
+      {:ok,
+       %{
+         profile: profile,
+         update_profile_gender_preference: genders
+       }} ->
```
ruslandoga, line 522:
```elixir
{:ok, changes} ->
%{...} = changes
```

lib/t/accounts.ex
```diff
@@ -488,48 +515,71 @@ defmodule T.Accounts do
...
-    Multi.run(multi, :update_profile_gender_preferences, fn _repo, %{profile: new_profile} ->
-      %Profile{filters: %Profile.Filters{genders: new_genders}} = new_profile
-
-      # TODO
-      new_genders = new_genders || []
+  defp update_profile_gender_preference(multi, profile, %{
+         "gender_preference" => gender_preference
+       }) do
+    update_profile_gender_preference(multi, profile, gender_preference)
+  end
```
ruslandoga, line 539:
```elixir
defp update_profile_gender_preference(multi, profile,attrs) when is_map(attrs) do
     update_profile_gender_preference(multi, profile, attrs[:gender_preference] || attrs["gender_preference"])
   end
```

lib/t/accounts.ex
```diff
@@ -488,48 +515,71 @@ defmodule T.Accounts do
...
-      %Profile{filters: %Profile.Filters{genders: new_genders}} = new_profile
-      insert_all_gender_preferences(new_genders || [], user_id)
-      {:ok, [to_add: new_genders, to_remove: []]}
+  defp update_profile_gender_preference(multi, %Profile{user_id: user_id}, _atrs) do
+    Multi.run(multi, :update_profile_gender_preference, fn _repo, _changes ->
+      genders =
+        GenderPreference
+        |> where(user_id: ^user_id)
+        |> select([g], g.gender)
+        |> Repo.all()
```
ruslandoga, line 572: this is not an update

ruslandoga: this clause should be noop

lib/t/accounts.ex
```diff
@@ -488,48 +515,71 @@ defmodule T.Accounts do
...
+         %Profile{user_id: user_id},
+         gender_preference
+       )
+       when is_list(gender_preference) do
+    Multi.run(multi, :update_profile_gender_preference, fn _repo, _changes ->
+      # TODO not delete the actual ones
       GenderPreference
       |> where(user_id: ^user_id)
-      |> where([p], p.gender in ^to_remove)
       |> Repo.delete_all()
```
ruslandoga, line 558: please keep it as before

lib/t/accounts.ex
```diff
@@ -547,6 +597,18 @@ defmodule T.Accounts do
     |> Repo.all()
   end
 
+  def get_feed_filter(user_id) do
```
ruslandoga, line 600: what's the point of this function? where is it called? why can't profile in memory be used to construct feed_filters?

lib/t/accounts/profile.ex
```diff
@@ -10,11 +10,11 @@ defmodule T.Accounts.Profile do
     field :story, {:array, :map}
     field :location, Geo.PostGIS.Geometry
 
-    # TODO remove
-    embeds_one :filters, Filters, primary_key: false, on_replace: :delete do
-      # ["F"] or ["F", "M"], etc.
-      field :genders, {:array, :string}
-    end
+    # filters
+    field :gender_preference, {:array, :map}, virtual: true
```
ruslandoga, line 14: it's not `{:array, :map}`, it's `{:array, :string}`

lib/t/accounts/profile.ex
```diff
@@ -58,17 +58,6 @@ defmodule T.Accounts.Profile do
         _ -> []
       end
     end)
-    |> cast_embed(:filters,
```
ruslandoga, line 61: this logic needs to be kept otherwise genders are not validated

lib/t/feeds.ex
```diff
@@ -115,42 +105,107 @@ defmodule T.Feeds do
     {feed_profiles, feed_cursor}
   end
 
-  defp continue_feed(user_id, location, gender, gender_preferences, count) do
+  defp continue_feed(user_id, location, gender, feed_filter, count) do
+    %{genders: gender_preferences, min_age: min_age, max_age: max_age, distance: distance} =
```
ruslandoga, line 109: `%FeedFilter{...} = `

lib/t/feeds.ex
```diff
@@ -115,42 +105,107 @@ defmodule T.Feeds do
...
+  end
+
+  defp maybe_apply_age_filters(query, min_age, max_age) do
+    query
+    |> maybe_apply_min_age_filer(min_age)
+    |> maybe_apply_max_age_filer(max_age)
+  end
+
+  defp maybe_apply_min_age_filer(query, min_age) do
+    case min_age do
```
ruslandoga, line 157:
```elixir
if min_age do
  where(...)
else
  query
end
```
anything more is pure complication

lib/t/feeds/feed_filter.ex
```diff
@@ -0,0 +1,4 @@
+defmodule T.Feeds.FeedFilter do
+  @enforce_keys [:genders]
```
ruslandoga, line 2: better to remove this

priv/repo/migrations/20211109083906_add_age_and_distance_preferences.exs
```diff
@@ -0,0 +1,16 @@
+defmodule T.Repo.Migrations.AddAgesAndDistance do
+  use Ecto.Migration
+
+  def change do
+    alter table(:profiles) do
+      add :min_age, :integer, default: 18, null: false
+      add :max_age, :integer, default: 100, null: false
+      add :distance, :integer, default: 20000, null: false
```
ruslandoga, line 8: remove defaults

priv/repo/migrations/20211109083906_add_age_and_distance_preferences.exs
```diff
@@ -0,0 +1,16 @@
...
+      remove :filters
+    end
+
+    create index(:profiles, [:min_age], where: "min_age is not null")
+    create index(:profiles, [:max_age], where: "max_age is not null")
+    create index(:profiles, [:distance], where: "distance is not null")
```
ruslandoga, line 14: where clause always returns true

test/t/feeds/feeds_test.exs
```diff
@@ -16,7 +16,7 @@ defmodule T.FeedsTest do
                  me.id,
                  me.profile.location,
                  _gender = "M",
-                 _gender_preference = ["F"],
+                 _feed_filter = %{genders: ["F"], min_age: nil, max_age: nil, distance: nil},
```
ruslandoga, line 19: feed filter needs to be a `%FeedFilter{}` struct, not plain map

ruslandoga (changes requested):

lib/t/accounts.ex
akhrail1996: It fetches feed_filers for feedChannel: https://github.com/getsince/test3/blob/8d4c6eb69d046f136ba2ddada861b46ccc4e224c/lib/t_web/channels/feed_channel.ex#L16

As socket assigns for different channels are separate, feedChannel doesn't have profile from memory. Alternatively the whole profile + gender_preferences could be assigned to feedChannel's socket

ruslandoga: if it's only for feed, than it should be in `Feeds` context

ruslandoga: `Accounts` are for working with users, profiles, auth, etc.

lib/t/feeds.ex
syfgkjasdkn: why not?

syfgkjasdkn: what I wrote above is easy to understand

ruslandoga: please do the approach above, it's easier to understand. applying `where` doesn't hurt, applying `where` only when there is a filter in db is correct

ruslandoga:
@akhrail1996 I'm a bit busy today, I think I'll be able to review it in the evening.

lib/t/accounts.ex
```diff
@@ -458,11 +469,17 @@ defmodule T.Accounts do
...
+      {:ok, changes} ->
+        case changes do
+          %{user: user, profile: %Profile{} = profile, gender_preferences: genders} ->
+            m = "user registered #{user.id}, user name is #{profile.name}"
+            Logger.warn(m)
+            Bot.async_post_message(m)
+            {:ok, %Profile{profile | gender_preference: genders, hidden?: false}}
+
+          %{profile: profile} ->
+            {:ok, %Profile{profile | hidden?: false}}
```
ruslandoga, line 481: why is this clause necessary?

lib/t/accounts.ex
```diff
@@ -488,22 +505,37 @@ defmodule T.Accounts do
...
     |> case do
-      {:ok, %{profile: profile}} -> {:ok, profile}
-      {:error, :profile, %Ecto.Changeset{} = changeset, _changes} -> {:error, changeset}
+      {:ok, changes} ->
+        case changes do
+          %{profile: profile, gender_preferences: genders} ->
+            {:ok, %Profile{profile | gender_preference: genders, hidden?: false}}
+
+          %{profile: profile} ->
+            {:ok, %Profile{profile | hidden?: false}}
```
ruslandoga, line 514: why `hidden?: false`? why is this clause necessary?

lib/t/accounts/profile.ex
```diff
@@ -58,17 +58,11 @@ defmodule T.Accounts.Profile do
...
-        |> validate_subset(:genders, @known_genders)
-        |> maybe_validate_required(opts, fn changeset ->
-          validate_required(changeset, [:genders])
-        end)
-      end
-    )
+    |> cast(attrs, [:gender_preference])
+    |> validate_subset(:gender_preference, @known_genders)
+    |> maybe_validate_required(opts, fn changeset ->
+      validate_required(changeset, [:gender_preference])
```
ruslandoga, line 64: duplicate` maybe_validate_required`

lib/t/feeds.ex
```diff
@@ -115,48 +105,114 @@ defmodule T.Feeds do
...
+  defp maybe_apply_age_filters(query, min_age, max_age) do
+    query
+    |> maybe_apply_min_age_filer(min_age)
+    |> maybe_apply_max_age_filer(max_age)
+  end
+
+  defp maybe_apply_min_age_filer(query, min_age) do
+    if min_age do
+      %{year: y, month: m, day: d} = DateTime.utc_now()
+      youngest = %Date{year: y - min_age, month: m, day: d}
```
ruslandoga, line 163: this might lead to invalid date (remember the leap year problem)

lib/t/feeds.ex
```diff
@@ -115,48 +105,114 @@ defmodule T.Feeds do
...
+      where(query, [p], p.birthdate >= ^oldest)
+    else
+      query
+    end
+  end
+
+  defp maybe_apply_distance_filter(query, location, distance) do
+    if distance do
+      meters = distance * 1000
+      where(query, [p], st_dwithin_in_meters(^location, p.location, ^meters) == true)
```
ruslandoga, line 185: `st_dwithin_in_meters` already return a bool

lib/t/feeds.ex
```diff
@@ -115,48 +105,114 @@ defmodule T.Feeds do
...
   end
 
+  def get_feed_filter(user_id) do
+    genders = T.Accounts.list_gender_preferences(user_id)
+
+    {min_age, max_age, distance} =
+      Profile
+      |> where(user_id: ^user_id)
+      |> select([p], {p.min_age, p.max_age, p.distance})
+      |> Repo.one()
```
ruslandoga, line 211: `Repo.one!`

lib/t_web/channels/feed_channel.ex
```diff
@@ -14,7 +13,7 @@ defmodule TWeb.FeedChannel do
       user_id = String.downcase(user_id)
       %{screen_width: screen_width} = socket.assigns
 
-      gender_preferences = Accounts.list_gender_preferences(user_id)
+      feed_filter = Feeds.get_feed_filter(user_id)
```
ruslandoga, line 16: how is it kept up-to-date with profile?

ruslandoga: feed channel can subscribe for profile updates and refetch feed filter in `handle_info` when the even arrives

priv/repo/migrations/20211109083906_add_age_and_distance_preferences.exs
```diff
@@ -0,0 +1,16 @@
...
+    alter table(:profiles) do
+      add :min_age, :integer
+      add :max_age, :integer
+      add :distance, :integer
+      remove :filters
+    end
+
+    create index(:profiles, [:min_age])
+    create index(:profiles, [:max_age])
+    create index(:profiles, [:distance])
```
ruslandoga, line 14: `where: "min_age is not null"` etc? since you removed `null: false`.

test/support/factory.ex
ruslandoga, line 113: since these are nullable, we shouldn't give them defaults in factory. otherwise it's impossible to test users without these fields

ruslandoga (changes requested):

lib/t/accounts.ex
akhrail1996: because there may be no :gender_preferences in this case: https://github.com/getsince/test3/blob/b94aebd7d5d4c04b4ef98b3054e7eaaccd8c9f01/lib/t/accounts.ex#L553

lib/t/accounts.ex
akhrail1996: it was there before, not sure why as well (maybe_unhide should probably be used here)

ruslandoga: it wasn't

lib/t/accounts.ex
ruslandoga: please use two clauses, current approach binds notification logic to ui, hard to understand

lib/t_web/channels/feed_channel.ex
akhrail1996: I just rejoin the feedChannel on client: https://github.com/getsince/test5/blob/rails-fork/Since/Feed/FeedCoordinator.swift#L207
but you're right, probably this approach is cleaner

lib/t/accounts.ex
ruslandoga: Screenshot 2021-11-13 at 18 49 33

akhrail1996: right, my bad

lib/t/accounts.ex
```diff
@@ -22,6 +22,21 @@ defmodule T.Accounts do
     AppleSignIn
   }
 
+  @pubsub T.PubSub
+  @topic "__a"
+
+  defp pubsub_user_topic(user_id) when is_binary(user_id) do
```
ruslandoga, line 28: I suggest subscribing for feed filters

lib/t/feeds.ex
```diff
@@ -115,48 +105,122 @@ defmodule T.Feeds do
...
+  end
+
+  defp date_years_from_now(years) do
+    %{year: y, month: m, day: d} = DateTime.utc_now()
+
+    case Date.new(y - years, m, d) do
+      {:ok, date} -> date
+      # leap year
+      {:error, :invalid_date} -> %Date{year: y - years, month: m, day: d - 1}
+    end
```
ruslandoga, line 167: no, I wanted something like `select '2020-02-29'::date - interval '1 year';`

lib/t_web/channels/feed_channel.ex
```diff
@@ -273,10 +272,13 @@ defmodule TWeb.FeedChannel do
     {:noreply, socket}
   end
 
-  # TODO refactor
-  defp render_feed_item(feed_item, screen_width) do
-    {%FeedProfile{} = profile, distance} = feed_item
-    assigns = [profile: profile, screen_width: screen_width, distance: distance]
+  def handle_info({Accounts, :feed_filter_updated, user_id}, socket) do
+    feed_filter = Feeds.get_feed_filter(user_id)
```
ruslandoga, line 276: this would result in a race condition with read replicas, I suggest broadcasting feed filters

lib/t/feeds.ex
ruslandoga:
```elixir
|> where([p], p.birthdate <= fragment("now() - ? * interval '1y'", min_age))
```
can be abstracted into a macro like distance_km

ruslandoga:
@akhrail1996 it's too tiresome to go back and forth ... if the the new functionality already works, please merge it without final review

akhrail1996:
> I'm afraid I would be finding something to fix in every change

https://www.youtube.com/watch?v=7PphbSFZWuU&ab_channel=MislavEljuga

seriously though I think at least I benefit from all these fixes, so thank you 🙏

PR #408, opened 2021-11-16
schedule push for registered users with empty stories

### LINKED ISSUE #405: send pushes to registered, but non-onboarded users (haven’t submit profile with a story)

akhrail1996:
closes #405 

- [x] manually send push to all current users with empty stories

### DIFF
--- a/lib/t/accounts.ex
+++ b/lib/t/accounts.ex
@@ -689,7 +689,7 @@ defmodule T.Accounts do
     Profile
     |> where([p], is_nil(p.story))
     |> where([p], p.last_active <= fragment("now() - interval '1 day'"))
-    |> where([p], p.last_active >= fragment("now() - interval '1 day 1 minute'"))
+    |> where([p], p.last_active > fragment("now() - interval '1 day 1 minute'"))
     |> select([p], p.user_id)
     |> Repo.all()
     |> Enum.map(fn user_id ->
--- a/lib/t/accounts.ex
+++ b/lib/t/accounts.ex
@@ -686,13 +686,10 @@ defmodule T.Accounts do
   end
 
   def push_users_to_complete_onboarding do
-    start_date = DateTime.add(DateTime.utc_now(), -24 * 60 * 60)
-    finish_date = DateTime.add(DateTime.utc_now(), -24 * 60 * 60 - 60)
-
     Profile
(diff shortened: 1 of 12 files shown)


lib/t/accounts.ex
```diff
@@ -682,4 +684,24 @@ defmodule T.Accounts do
...
   end
+
+  def push_users_to_complete_onboarding do
+    start_date = DateTime.add(DateTime.utc_now(), -24 * 60 * 60)
+    finish_date = DateTime.add(DateTime.utc_now(), -24 * 60 * 60 - 60)
+
+    Profile
+    |> where([p], is_nil(p.story))
+    |> where([p], p.last_active <= ^start_date)
+    |> where([p], p.last_active >= ^finish_date)
```
akhrail1996, line 695: didn't figure how to use SQL fragment now() with correct timezone

ruslandoga:
I don't understand the goal here

lib/t/accounts.ex
ruslandoga: what timezone is correct?

akhrail1996: UTC?

ruslandoga:
```sql
select now();
-- 2021-11-16 11:25:40.825731+00
```

```elixir
iex(1)> DateTime.utc_now 
~U[2021-11-16 11:25:48.669111Z]
```

akhrail1996:
The goal is to "activate" users which started profile creation, but switched to other things / forgot about it at some point and never finished their stories -> never actually tried the app

surprisingly there is a lot of such users rn (181 / 530)

lib/t/accounts.ex
ruslandoga: https://xyproblem.info

ruslandoga:
how does it work?

lib/t/accounts.ex
akhrail1996: thanks, the solution to this indeed turned out to be:
`psql`
`ALTER DATABASE t_test SET timezone TO 'UTC';`

akhrail1996: @orlovvlad95 once I merge this PR into master you will experience this issue as well when testing, so pls run the above commands in your shell

PR #412, opened 2021-11-20
bugfix

orlovvlad95:


lib/t/push_notifications/dispatch_job.ex
```diff
@@ -54,8 +53,17 @@ defmodule T.PushNotifications.DispatchJob do
...
-      if profile = profile_info(receiver_id) do
+    if match = alive_match(match_id) do
+      %Matches.Match{user_id_1: uid1, user_id_2: uid2} = match
+
+      picker_id =
+        if(receiver_id == uid1) do
+          uid2
+        else
+          uid1
+        end
```
akhrail1996, line 64: I would add `picker_id` to the `args` above to declutter the code here
can be done in places like this:

lib/t/push_notifications/dispatch_job.ex
```diff
@@ -73,15 +81,28 @@ defmodule T.PushNotifications.DispatchJob do
...
 
-        receiver_id
-        |> Accounts.list_apns_devices()
-        |> schedule_apns(type, %{"match_id" => match_id})
+        picker_id =
+          if(receiver_id == uid1) do
+            uid2
+          else
+            uid1
+          end
```
akhrail1996, line 97: same here and everywhere below

akhrail1996: this push was not in the last PR at all, not good

akhrail1996 (changes requested):
I've deployed this fix, but it needs some polishing before merging

On a separate note, the pushes' titles we have now do not fit even on the largest screens (12ProMax), let alone smaller ones. Adding names only makes things worse, especially if the name is a long one. Also if a name is closer to the end of the push it may not be displayed as well.

image

So I think we need to make push titles short in general and keep the name in the titles. Where this is not possible, we can move names to body of pushes.

lib/t/push_notifications/dispatch_job.ex
orlovvlad95: good idea, made like this

lib/t/push_notifications/dispatch_job.ex
orlovvlad95: yeah, decide to add too

orlovvlad95:
> I've deployed this fix, but it needs some polishing before merging
> 
> On a separate note, the pushes' titles we have now do not fit even on the largest screens (12ProMax), let alone smaller ones. Adding names only makes things worse, especially if the name is a long one. Also if a name is closer to the end of the push it may not be displayed as well.
> 
> image
> 
> So I think we need to make push titles short in general and keep the name in the titles. Where this is not possible, we can move names to body of pushes.

@akhrail1996 I edit some pushes text, pls check my version of them

lib/t/push_notifications/dispatch_job.ex
```diff
@@ -51,18 +51,9 @@ defmodule T.PushNotifications.DispatchJob do
...
-    if match = alive_match(match_id) do
-      %Matches.Match{user_id_1: uid1, user_id_2: uid2} = match
-
-      picker_id =
-        if(receiver_id == uid1) do
-          uid2
-        else
-          uid1
-        end
+    %{"match_id" => match_id, "receiver_id" => receiver_id, "picker_id" => picker_id} = args
```
akhrail1996, line 54: this function is designed for two types of pushes: `["timeslot_offer", "timeslot_accepted"]`

`"timeslot_offer"` push doesn't have `"picker_id"` in its args: 
image

therefore this line would crash. please be more attentive and double-check your work

akhrail1996:
> @akhrail1996 I edit some pushes text, pls check my version of them

This (one of the most problematic)
image
can be changed to sth like "Timur invites you on a date"

orlovvlad95:
> > @akhrail1996 I edit some pushes text, pls check my version of them
> 
> This (one of the most problematic) image can be changed to sth like "Timur invites you on a date".
> 
> The others look good 👍

@akhrail1996 changed

lib/t/push_notifications/dispatch_job.ex
orlovvlad95: ok, will do, sorry
add picker

priv/gettext/en/LC_MESSAGES/apns.po
```diff
@@ -64,7 +64,7 @@ msgstr "Come in and call now 👉"
 #, elixir-format
 #: lib/t/push_notifications/apns.ex:109
 msgid "%{name} пригласил%{gender_a} тебя на дэйт!"
-msgstr "You've been asked out on a date by %{name}!"
+msgstr "%{name} invites you on a date!"
```
akhrail1996, line 67: when you edit push, you are supposed to change it in minimum like 4 places:
* dgettext (where the push is created)
* apns.po (master file of texts of all pushes)
* ru apns.pot
* en apns.pot

the way you should change pushes is to first change it just in code (dgettext), then run `mix gettext.extract --merge` , which will automatically trigger changes in all the other 3 places, and then if necessary correct the translations in relevant .pot files.

akhrail1996 (approved):

PR #426, opened 2021-11-29
Pushes after call has started

orlovvlad95:


lib/t/matches.ex
```diff
@@ -894,4 +895,65 @@ defmodule T.Matches do
     )
     |> where([m, e, c], is_nil(c.timestamp))
   end
+
+  def check_matches_calls() do
+    matches_without_successful_calls =
```
akhrail1996, line 900: name of the query does not match its content

lib/t/matches.ex
```diff
@@ -894,4 +895,65 @@ defmodule T.Matches do
...
+      Match
+      |> join(:inner, [m], c in Call,
+        on:
+          (c.caller_id == m.user_id_1 and c.called_id == m.user_id_2) or
+            (c.caller_id == m.user_id_2 and c.called_id == m.user_id_1)
+      )
+      |> where([m, c], c.accepted_at >= fragment("now() - interval '80 hours'"))
+      |> select([m], m.id)
+
+    matches_for_push =
```
akhrail1996, line 910: same here, not all the matches inside this query will receive a push (e.g. matches with successful calls will be included here as well, but we won't send them any pushes)

lib/t/matches.ex
```diff
@@ -894,4 +895,65 @@ defmodule T.Matches do
...
+      |> where([t], t.selected_slot >= fragment("now() - interval '80 hours'"))
+      |> where([t], t.match_id not in subquery(matches_without_successful_calls))
+      |> select([t], t.match_id)
+
+    good_users =
+      Call
+      |> join(:inner, [c], m in Match,
+        on:
+          (c.caller_id == m.user_id_1 and c.called_id == m.user_id_2) or
+            (c.caller_id == m.user_id_2 and c.called_id == m.user_id_1)
```
akhrail1996, line 921: all of these is super messy
if needed, just add match_id column to calls table (just keep in mind that there is an edge case of user calling back from a "missed call" when they are already not matched with the mater -> match_id will be nil here)

lib/t/matches.ex
```diff
@@ -894,4 +895,65 @@ defmodule T.Matches do
...
+      )
+      |> where([c, m], c.inserted_at >= fragment("now() - interval '80 hours'"))
+      |> where([c, m], m.id in subquery(matches_for_push))
+      |> where([c, m], c.caller_id == m.user_id_1 or c.caller_id == m.user_id_2)
+      |> select([c, m], c.caller_id)
+      |> T.Repo.all()
+
+    not_in_good_users =
```
akhrail1996, line 929: fail to see how is this different from good_users above

lib/t/matches.ex
```diff
@@ -894,4 +895,65 @@ defmodule T.Matches do
...
+      Call
+      |> join(:inner, [c], m in Match,
+        on:
+          (c.caller_id == m.user_id_1 and c.called_id == m.user_id_2) or
+            (c.caller_id == m.user_id_2 and c.called_id == m.user_id_1)
+      )
+      |> where([c, m], c.inserted_at >= fragment("now() - interval '80 hours'"))
+      |> where([c, m], m.id in subquery(matches_for_push))
+      |> where([c, m], c.called_id == m.user_id_1 or c.called_id == m.user_id_2)
+      |> where([c, m], c.called_id not in subquery(not_in_good_users))
```
akhrail1996, line 951: if `not_in_good_users` only used here it can be substituted with `not in ^good_users`

PR #439, opened 2021-12-16
Newbies Since Live + Scheduling moved to Oban

ruslandoga:
Adds a new type of Since Live: "Since Live for Newbies".

"Since Live for Newbies" is meant to cut the time of the first call for the newly registered users up to most a day.
"Since Live for Newbies" participants are these newly registered users (newbies) and some hardcoded "old" users who could have calls with newbies and introduce them to the app. Newbies can call other newbies as well.

This PR updates feed channel to support "Since Live for Newbies" and also adds scheduling logic (using Oban) to notify about and start on time "Since Live for Newbies" events.


config/config.exs
```diff
@@ -3,6 +3,8 @@ import Config
 config :t, ecto_repos: [T.Repo]
 config :t, T.Repo, types: T.PostgresTypes
 
+config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
```
ruslandoga, line 6: To be able to shift timezones in `DateTime` functions

lib/t/application.ex
```diff
@@ -29,6 +29,30 @@ defmodule T.Application do
         unless_disabled(T.Matches.MatchExpirer),
         unless_disabled(T.PushNotifications.ScheduledPushes),
         unless_disabled(T.Feeds.LiveModeManager),
+        unless_disabled(
+          {T.Feeds.NoobsLiveModeManager,
+           schedule: [
+             %{
+               at: {~T[13:00:00], "Europe/Moscow"},
+               run: {T.Feeds, :notify_noobs_live_mode_will_be_today, []}
```
ruslandoga, line 37: these functions are not implemented yet

lib/t/feeds.ex
```diff
@@ -41,15 +41,15 @@ defmodule T.Feeds do
   end
 
   defp notify_subscribers(:mode_change, event) do
-    broadcast_from(mode_change_topic(), {__MODULE__, [:mode_change, event]})
+    broadcast(mode_change_topic(), {__MODULE__, [:mode_change, event]})
```
ruslandoga, line 44: As this `notify_subscribers(:mode_change, event)` is called from a non-channel process, I think it's ok to `broadcast` it instead of `broadcast_from` since the latter is mostly useful when we don't want the broadcasted to receive their own broadcast.

lib/t/feeds.ex
```diff
@@ -625,4 +629,148 @@ defmodule T.Feeds do
...
+    Bot.async_post_message(m)
+
+    # we don't need to send `end` event only to newbies,
+    # we can turn off everyone who's live right now
+    # since "newbie live" and "normal live" don't intersect
+    notify_subscribers(:mode_change, :end)
+
+    two_hours_ago_ref = DateTime.add(DateTime.utc_now(), -7200)
+    clear_live_tables(two_hours_ago_ref)
+
```
ruslandoga, line 686: Not sure about what push notifications to send when the "newbies" version of Since Live ends. In the real Since Live the notifications say when the next Since Live is, should it be the same here, it would say when the next real Since Live is?

lib/t/feeds.ex
```diff
@@ -625,4 +629,148 @@ defmodule T.Feeds do
...
+    :ok
+  end
+
+  @doc """
+  Lists ids of today's "newbies live" participants.
+
+  "newbies live" participants are:
+  - new newbies: users who have registered today (they are still newbies even if they've had a call)
+  - TODO old newbies: users who have registered before but haven't had a call
```
ruslandoga, line 695: I want to do this in the next PR, does it make sense to include users who have registered some time ago but haven't had a call into "newbies" Since Live?

lib/t/feeds.ex
```diff
@@ -625,4 +629,148 @@ defmodule T.Feeds do
...
+  @spec newbies_list_today_participants :: [Ecto.UUID.t()]
+  def newbies_list_today_participants(reference \\ DateTime.utc_now()) do
+    # maybe Repo.stream later
+    # newbies = new_newbies_q(reference) |> union(^old_newbies_q()) |> Repo.all()
+    newbies = new_newbies_q(reference) |> Repo.all()
+    hardcoded_oldies() ++ newbies
+  end
+
+  defp hardcoded_oldies do
+    Application.get_env(:t, :oldies) || []
```
ruslandoga, line 707: We'd need to decide who'd be an oldie. I guess we either hardcode it into `config/runtime.exs` or pass via env var from nomad. `OLDIES=0000017d-d720-1f6a-06e9-e8bbc6570000,0000017d-d720-1f6a-06e9-e8bbc6570000`

lib/t/feeds.ex
```diff
@@ -625,4 +629,148 @@ defmodule T.Feeds do
...
+
+  Assumes "Since Live" event starts at 19:00 and ends at 21:00 MSK.
+  """
+  @spec newbies_live_now?(Ecto.UUID.t(), DateTime.t()) :: boolean
+  def newbies_live_now?(user_id, reference \\ DateTime.utc_now()) do
+    msk_now = DateTime.shift_zone!(reference, "Europe/Moscow")
+
+    newbie_time? =
+      Time.compare(~T[19:00:00], msk_now) in [:eq, :lt] and
+        Time.compare(~T[21:00:00], msk_now) in [:eq, :gt]
```
ruslandoga, line 768: I'm assuming Since Live for newbies would be as long as the real one, 2 hours, and that it starts at 19:00 MSK.

lib/t/push_notifications/apns.ex
```diff
@@ -280,6 +280,41 @@ defmodule T.PushNotifications.APNS do
...
+  end
+
+  def build_alert_payload("newbie_live_mode_started" = type, _data) do
+    alert = %{
+      "title" => dgettext("apns", "Since Live starts 🥳"),
+      "body" => dgettext("apns", "Only you and other new users are invited 🎉")
+    }
+
+    base_alert_payload(type, alert, %{})
+  end
```
ruslandoga, line 316: I guess the push notification massages need to be updates slightly to mention that it's for newbies. I've updated only the `newbie_live_mode_started` one, but it's not good. I'll think a bit and post some variants here.

lib/t/push_notifications/dispatch_job.ex
```diff
@@ -296,6 +296,15 @@ defmodule T.PushNotifications.DispatchJob do
     :ok
   end
 
+  defp handle_type(type, _args)
+       when type in ["newbie_live_mode_today", "newbie_live_mode_soon"] do
+    Feeds.newbies_list_today_participants()
```
ruslandoga, line 301: Oops, that's bad since when it runs at 13:00 (for `newbie_live_mode_today`) it would include yesterdays newbies. I'll need to fix it.

lib/t/push_notifications/dispatch_job.ex
```diff
@@ -326,6 +335,7 @@ defmodule T.PushNotifications.DispatchJob do
         "priority" => "10"
       })
     end)
+    # TODO might need to chunk later (pg won't accept more than ~30k uuids at a time)
```
ruslandoga, line 338: Just a general TODO about postgres' insert payload max size

lib/t_web/channels/feed_channel.ex
```diff
@@ -407,10 +408,19 @@ defmodule TWeb.FeedChannel do
...
+    socket =
+      case {event, socket.assigns[:mode]} do
+        {:start, "normal"} ->
+          push(socket, "live_mode_started", %{})
+          assign(socket, mode: "live")
+
+        {:end, "live"} ->
+          push(socket, "live_mode_ended", %{})
+          assign(socket, mode: "normal")
+
```
ruslandoga, line 420: `{event, ...}` is to avoid sending `live_mode_ended` to users who weren't even in the live event, for example, when newbies "live" ends, all `:mode_change` subscribers receive `[:mode_change, :end]` but only the newbies react to it (since only they are `"live"` in it).

mix.exs
```diff
@@ -74,7 +74,8 @@ defmodule T.MixProject do
       {:cloud_watch, github: "getsince/cloud_watch", branch: "drop-httpoison"},
       {:benchee, "~> 1.0", only: :bench},
       {:dialyxir, "~> 1.1", only: :dev, runtime: false},
-      {:jose, "~> 1.11"}
+      {:jose, "~> 1.11"},
+      {:tzdata, "~> 1.1.1"}
```
ruslandoga, line 78: https://hexdocs.pm/tzdata/readme.html

priv/gettext/ru/LC_MESSAGES/apns.po
```diff
@@ -180,12 +180,17 @@ msgstr "Приходи на вечеринку к 19:00, будет классн
...
 #, elixir-format
-#: lib/t/push_notifications/apns.ex:241
+#: lib/t/push_notifications/apns.ex:241 lib/t/push_notifications/apns.ex:293
 msgid "Since LIVE today 🥳"
 msgstr "Since LIVE уже сегодня 🥳"
+
+#, elixir-format
+#: lib/t/push_notifications/apns.ex:312
+msgid "Only you and other new users are invited 🎉"
+msgstr "Приглашены все новички 🎉"
```
ruslandoga, line 196: Translations and stuff, good enough?

config/runtime.exs
```diff
@@ -26,9 +26,46 @@ config :sentry,
 
 config :t, T.PromEx, disabled: config_env() != :prod
 
+crontab =
+  case config_env() do
+    :prod ->
+      alias T.PushNotifications.DispatchJob
+
+      [
+        if System.get_env("ENABLE_NEWBIE_LIVE") do
```
akhrail1996, line 35: gotta not forget to add this to .env

config/runtime.exs
```diff
@@ -26,9 +26,46 @@ config :sentry,
...
+            # At 13:00 on every day-of-week from Sunday through Thursday and Saturday.
+            {"0 13 * * 0-4,6", DispatchJob, args: %{"type" => "newbie_live_mode_today"}},
+            # https://crontab.guru/#45_18_*_*_0-4,6
+            # At 18:45 on every day-of-week from Sunday through Thursday and Saturday.
+            {"45 18 * * 0-4,6", DispatchJob, args: %{"type" => "newbie_live_mode_soon"}},
+            # https://crontab.guru/#00_19_*_*_0-4,6
+            # At 19:00 on every day-of-week from Sunday through Thursday and Saturday.
+            {"00 19 * * 0-4,6", T.Feeds.NewbiesLive.StartJob},
+            # https://crontab.guru/#00_20_*_*_0-4,6
+            # At 21:00 on every day-of-week from Sunday through Thursday and Saturday.
```
akhrail1996, line 47: I think one hour is enough for newbie live

config/runtime.exs
```diff
@@ -259,6 +296,9 @@ if config_env() == :test do
 
   config :t, T.Feeds.SeenPruner, disabled?: true
   config :t, T.Feeds.LiveModeManager, disabled?: true
+
+  # there is no user with this id, it's been generated just for the tests
+  config :t, oldies: ["0000017d-d720-1f6a-06e9-e8bbc6570000"]
```
akhrail1996, line 301: we can start with the following "oldies":
Vlad: 0000017b-b0cf-d3e8-0242-ac1100020000
Lily: 0000017b-e3d2-2b30-0242-ac1100020000
me: 0000017c-14c7-9745-0242-ac1100020000

lib/t/feeds.ex
```diff
@@ -625,4 +629,149 @@ defmodule T.Feeds do
...
+
+  This function shouldn't be called directly, but rather scheduled with Oban in crontab.
+  """
+  @spec newbies_end_live :: :ok
+  def newbies_end_live do
+    m = "LIVE (newbie's special edition) ended"
+    Logger.warn(m)
+    Bot.async_post_message(m)
+
+    # we don't need to send `end` event only to newbies,
```
akhrail1996, line 679: ~~would break current frontend: it will change the mode for non-newbies
https://github.com/getsince/test5/blob/rails-fork/Since/Feeds/Live%20Feed/LiveFeedChannel.swift#L60
so better send this event only to newbies

(though I recognize that this frontend "feature" should be fixed)~~

lib/t/feeds.ex
akhrail1996: > should it be the same here, it would say when the next real Since Live is?

I think so

lib/t/feeds.ex
akhrail1996: nope, and I wouldn't bother users with no call with newbies LIVE everyday (at this point I accept the fact that some people need time to "mature" for audio-dates)

lib/t/feeds.ex
akhrail1996: I would prefer hardcoding into runtime.exs

lib/t/feeds.ex
```diff
@@ -625,4 +629,149 @@ defmodule T.Feeds do
...
+  end
+
+  # TODO move to accounts.ex? maybe later
+  @spec new_newbies_q(DateTime.t()) :: Ecto.Query.t()
+  defp new_newbies_q(reference) do
+    T.Accounts.User
+    |> where([u], is_nil(u.blocked_at))
+    |> where([u], not is_nil(u.onboarded_at))
+    |> where([u], fragment("? + interval '24h' >= ?::timestamp ", u.onboarded_at, ^reference))
+    |> join(:inner, [u], p in FeedProfile, on: u.id == p.user_id and not p.hidden?)
```
akhrail1996, line 717: this fixes the issue with inviting users who didn't complete their stories, but now these users will never be invited to newbies LIVE
this maybe not so critical, but maybe fixed by introducing new column to users table (~~"real" onboarded.at~~, "onboarded_with_profile_at")?

lib/t/feeds.ex
akhrail1996: I would say 1 hour is enough

lib/t/push_notifications/apns.ex
```diff
@@ -280,6 +280,41 @@ defmodule T.PushNotifications.APNS do
     base_alert_payload(type, alert, %{})
   end
 
+  # newbies alerts
+
+  # maybe have some play on `orientation` (as in uni)
```
akhrail1996, line 285: I will ask Lily to write special push texts 👍

lib/t_web/channels/feed_channel.ex
akhrail1996: ah, this seem to solve my earlier comment above

priv/gettext/ru/LC_MESSAGES/apns.po
akhrail1996: Lily will refactor if needed ;)

lib/t/push_notifications/dispatch_job.ex
ruslandoga: fixed

lib/t/feeds.ex
ruslandoga: I've opened a separate PR for this: https://github.com/getsince/test3/pull/441

test/t/feeds/newbies_live_test.exs
```diff
@@ -0,0 +1,220 @@
...
+             ] = Enum.map(all_enqueued(), & &1.args)
+
+      expected_alerts = %{
+        "BABABABABA" => %{
+          "title" => "Since Live for babies is over ✌️",
+          "body" => "Wait for the real party on Monday 👀"
+        },
+        "ABABABAB" => %{
+          "title" => "Since Live для малышей закончился ✌️",
+          "body" => "Жди следующую вечеринку в понедельник 👀"
```
ruslandoga, line 191: - [ ] next real Since Live would be on Thursday, actually ...

ruslandoga: handled in https://github.com/getsince/test3/pull/445

priv/gettext/ru/LC_MESSAGES/apns.po
ruslandoga: @akhrail1996 I think this is what's left

config/runtime.exs
ruslandoga: and this

lib/t/push_notifications/apns.ex
```diff
@@ -228,18 +228,10 @@ defmodule T.PushNotifications.APNS do
...
-      case day_of_week do
-        6 -> dgettext("apns", "Come to the party at 20:00, it will be 🔥")
-        4 -> dgettext("apns", "Come to the party at 19:00, it will be 🔥")
-      end
-
+  def build_alert_payload("live_mode_today" = type, %{"time" => time}) do
     alert = %{
       "title" => dgettext("apns", "Since LIVE today 🥳"),
-      "body" => body
+      "body" => dgettext("apns", "Come to the party at %{time}, it will be 🔥", time: time)
```
akhrail1996, line 234: %{time} here is for the core TZ? so for user in the same region, but different TZ it will be misleading, right?

akhrail1996 (approved):

lib/t/push_notifications/apns.ex
ruslandoga: moved to #461

PR #442, opened 2021-12-20
set users.onboarded_with_story_at during onboarding

ruslandoga:
continuing from https://github.com/getsince/test3/pull/441 because I forgot to do it there ...

### DIFF
--- a/lib/t/accounts.ex
+++ b/lib/t/accounts.ex
@@ -523,6 +523,11 @@ defmodule T.Accounts do
     end)
     |> Multi.run(:show_profile, fn _repo, %{user: user} ->
       {count, nil} = maybe_unhide_profile_with_story(user.id)
+
+      if count >= 1 do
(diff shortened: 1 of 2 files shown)


lib/t/accounts.ex
```diff
@@ -521,8 +521,17 @@ defmodule T.Accounts do
...
       {count, nil} = maybe_unhide_profile_with_story(user.id)
+
+      # TODO move someplace else
+      if count >= 1 do
+        User
+        |> where(id: ^user.id)
+        |> update([u], set: [onboarded_with_story_at: fragment("now()")])
+        |> repo.update_all([])
+      end
+
```
ruslandoga, line 534: not nice, but it gets the job done, sets `onboarded_with_story_at` first time the profile's `hidden?` toggles

akhrail1996 (approved):

PR #445, opened 2021-12-21
consolidate live events scheduling

ruslandoga:
This PR tries to consolidate both types of Sinve Live events (real and newbies) into a "coherent" whole with a single place to configure both.

Continues #439 

- [x] single place to manage Since Live schedule
- [x] Since Live and Since Live for newbies use the same schedule and work together
- [x] don't let MSK tz leak from Feeds module, always convert to Etc/UTC
- [x] tests!
- [x] cleanup!
- [x] docs!


lib/t/feeds.ex
```diff
@@ -633,8 +633,181 @@ defmodule T.Feeds do
...
+    [
+      {_mon = 1, :newbie, [newbies_live_start_at(), newbies_live_end_at()]},
+      {_tue = 2, :newbie, [newbies_live_start_at(), newbies_live_end_at()]},
+      {_wed = 3, :newbie, [newbies_live_start_at(), newbies_live_end_at()]},
+      {_thu = 4, :real, [_start = ~T[19:00:00], _end = ~T[21:00:00]]},
+      {_fri = 5, :newbie, [newbies_live_start_at(), newbies_live_end_at()]},
+      {_sat = 6, :real, [_start = ~T[18:00:00], _end = ~T[20:00:00]]},
+      {_sun = 7, :newbie, [newbies_live_start_at(), newbies_live_end_at()]}
+    ]
+  end
```
ruslandoga, line 648: This would be the place to configure them.

lib/t/feeds.ex
```diff
@@ -633,8 +633,181 @@ defmodule T.Feeds do
...
+
+      reference
+      |> Date.add(diff)
+      |> DateTime.new!(start, "Europe/Moscow")
+    end)
+    |> Enum.sort({:asc, Date})
+    |> Enum.find(fn datetime ->
+      DateTime.compare(datetime, reference) in [:gt, :eq]
+    end)
+  end
```
ruslandoga, line 695: This function can replace `since_live_date`

ruslandoga: @akhrail1996 this function returns the datetime for the next real live event. I wonder what should be return when it's called during an ongoing live event? Like here:

```elixir
# if today's event has started, next event is returned
iex> thu_19_00_01 = DateTime.new!(~D[2021-12-23], ~T[19:00:01], "Europe/Moscow")
iex> live_next_real_at(thu_19_00_01)
_sat_18_00 = DateTime.new!(~D[2021-12-25], ~T[18:00:00], "Europe/Moscow")
```

Maybe if the event has started but hasn't ended, current event start datetime is returned, but if it has ended, next event datetime is returned?

lib/t/application.ex
```diff
@@ -25,10 +25,9 @@ defmodule T.Application do
         TWeb.Presence,
         TWeb.UserSocket.Monitor,
         T.Repo,
-        unless_disabled(T.Feeds.SeenPruner),
-        unless_disabled(T.Matches.MatchExpirer),
-        unless_disabled(T.PushNotifications.ScheduledPushes),
-        unless_disabled(T.Feeds.LiveModeManager),
+        # unless_disabled(T.Feeds.SeenPruner),
+        # unless_disabled(T.Matches.MatchExpirer),
+        # unless_disabled(T.PushNotifications.ScheduledPushes),
```
ruslandoga, line 30: this is temporary

lib/t/feeds.ex
akhrail1996: yeah

lib/t/feeds.ex
akhrail1996: > Maybe if the event has started but hasn't ended, current event start datetime is returned, but if it has ended, next event datetime is returned?

yep

config/runtime.exs
```diff
@@ -252,7 +235,12 @@ if config_env() == :dev do
     static_bucket: System.fetch_env!("AWS_S3_BUCKET_STATIC"),
     static_cdn: System.fetch_env!("STATIC_CDN")
 
-  config :t, T.Media.Static, disabled?: false
+  config :t, T.Media.Static, disabled?: !!System.get_env("DISABLE_MEDIA")
+  config :t, T.Feeds.SeenPruner, disabled?: !!System.get_env("DISABLE_SEEN_PRUNER")
+  config :t, T.Matches.MatchExpirer, disabled?: !!System.get_env("DISABLE_MATCH_EXPIRER")
+
+  config :t, T.PushNotifications.ScheduledPushes,
+    disabled?: !!System.get_env("DISABLE_SCHEDULED_PUSHES")
```
ruslandoga, line 243: These are not disabled by default. And they are disabled if a relevant env var is set to anything `export DISABLE_MEDIA=1`.

lib/t/application.ex
```diff
@@ -28,7 +28,6 @@ defmodule T.Application do
         unless_disabled(T.Feeds.SeenPruner),
         unless_disabled(T.Matches.MatchExpirer),
         unless_disabled(T.PushNotifications.ScheduledPushes),
-        unless_disabled(T.Feeds.LiveModeManager),
```
ruslandoga, line 31: `T.Feeds.LiveModeManager` has been replaced with oban's cron jobs.

config/runtime.exs
ruslandoga: I disabled all these background workers to avoid accidentally sending someone a push notification or something like that (although it shouldn't happen since I'm using sandboxed creds to apns).

lib/t/calls.ex
```diff
@@ -64,8 +65,8 @@ defmodule T.Calls do
     |> Repo.exists?()
   end
 
-  defp live_mode?() do
-    T.Feeds.is_now_live_mode()
+  defp in_live_mode?(caller_id, called_id, reference) do
+    Feeds.live_now?(caller_id, reference) and Feeds.live_now?(called_id, reference)
```
akhrail1996, line 69: potential #TODO - do not allow calls for blocked / reported
i.e. during LIVE:
A invites B
B calls A
B harrases A, A ends the call & reports B
B still can call A

priv/gettext/ru/LC_MESSAGES/apns.po
```diff
@@ -166,76 +156,76 @@ msgid "Мэтч %{name} онлайн 👋"
...
 #, elixir-format
-#: lib/t/push_notifications/apns.ex:312
+#: lib/t/push_notifications/apns.ex:297
 msgid "Only you and other new users are invited 🎉"
 msgstr "Приглашены все новички 🎉"
 
 #, elixir-format
-#: lib/t/push_notifications/apns.ex:333
+#: lib/t/push_notifications/apns.ex:305
 msgid "Since Live for babies is over ✌️"
```
akhrail1996, line 180: gotta change to "newbies" (in case we forget to properly refactor these)

akhrail1996 (approved):
👍

lib/t/calls.ex
ruslandoga: moved to https://github.com/getsince/test3/issues/447

PR #449, opened 2021-12-23
don't ignore like_ratio in factory

ruslandoga:
closes https://github.com/getsince/test3/issues/448


test/support/factory.ex
```diff
@@ -125,15 +125,14 @@ defmodule T.Factory do
       birthdate: opts[:birthdate] || "1998-10-28",
       gender: gender,
       name: opts[:name] || "that",
-      like_ratio: opts[:like_ratio] || 0,
```
ruslandoga, line 128: `like_ratio` was ignored by `Accounts.onboard_profile/2`

akhrail1996 (approved):

PR #452, opened 2021-12-24
upgrade elixir, erlang, and alpine

ruslandoga:


mix.exs
```diff
@@ -5,7 +5,7 @@ defmodule T.MixProject do
     [
       app: :t,
       version: "0.1.6",
-      elixir: "~> 1.12",
+      elixir: "~> 1.13",
```
ruslandoga, line 8: @akhrail1996 what version do you have installer `elixir -v`?

akhrail1996 (approved):

PR #455, opened 2021-12-24
live pubsub gender filter

### LINKED ISSUE #178: optimise gender preference filters lookup when building feed

ruslandoga:
closes #178 

This PR adds user id, gender, and gender preferences as metadata to the `live` topic. When another user (say, Matty) activates their session, only the users with matching preferences are sent the message. In addition, users who have blocked or have been blocked by Matty are not sent any message. And users who are matched with Matty are sent a special "match online" message instead of the generic one.

### DIFF
--- a/lib/t/feeds.ex
+++ b/lib/t/feeds.ex
@@ -25,9 +25,6 @@ defmodule T.Feeds do
   @pubsub T.PubSub
   @topic "__f"
 
-  def subscribe_for_live_sessions, do: subscribe(live_topic())
-  defp live_topic, do: "__live:"
-
   def subscribe_for_mode_change, do: subscribe(mode_change_topic())
   defp mode_change_topic, do: "__mode_change:"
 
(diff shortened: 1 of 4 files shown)


test/t_web/channels/feed_channel_live_test.exs
```diff
@@ -283,44 +283,46 @@ defmodule TWeb.FeedChanneLiveTest do
       we = onboarded_user(gender: "M", accept_genders: ["F", "N"])
       joined_mate(%{mate: we})
 
-      p1 = onboarded_user(gender: "F", accept_genders: ["F", "N", "M"])
+      %{id: user_id} = p1 = onboarded_user(gender: "F", accept_genders: ["F", "N", "M"])
       joined_mate(%{mate: p1})
 
-      assert_push("activated_profile", _profile)
+      assert_push "activated_profile", %{"profile" => %{user_id: ^user_id}}
```
ruslandoga, line 289: These tests used to pass when they shouldn't have when run concurrently with tests in `feeds_live_tests.exs`, so I added a pattern with user id to match.

ruslandoga: In `feeds_live_tests.exs` there were sessions being activated with users without stories, which were broadcasted to out channel here making the channel crash.

test/t/feeds/feeds_live_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule T.FeedsLiveTest do
-  use T.DataCase, async: true
```
ruslandoga, line 2: This tests interfered with `feed_channel_live_test.exs` so I made it run separately.

akhrail1996 (approved):
elegant solution 👍🙏

lib/t/feeds.ex
```diff
@@ -376,49 +353,100 @@ defmodule T.Feeds do
...
   end
 
-  defp maybe_schedule_push_to_matches(user_id) do
-    Match
-    |> where([m], m.user_id_1 == ^user_id or m.user_id_2 == ^user_id)
-    |> select([m], {m.user_id_1, m.user_id_2})
-    |> Repo.all()
-    |> Enum.map(fn {user_id_1, user_id_2} ->
-      [mate_id] = [user_id_1, user_id_2] -- [user_id]
+  defp gendered_live_topic, do: "__live"
```
ruslandoga, line 413: As this is not a "gendered" topic anymore, I'll rename it to `live_topic` and review the PR one more time and set the "merger" to be you, @akhrail1996, so that you'd merge it when appropriate.

PR #462, opened 2021-12-24
lazy load /admin/profiles

ruslandoga:
small "quality of life" improvement: don't load all profiles at once at 800px resolution on `/admin/profiles`, but instead load them progressively as needed at `250px` resolution

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -39,6 +39,49 @@ const ScrollWindowDownHook = {
   },
 };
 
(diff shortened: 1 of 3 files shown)


akhrail1996:
@ruslandoga doesn't seem to work well locally: https://www.loom.com/share/10f8200d0cb8474d98f2e1c25a6a36fd

akhrail1996 (approved):

PR #469, opened 2021-12-24
match users after live call

ruslandoga:


lib/t/calls.ex
```diff
@@ -213,6 +213,10 @@ defmodule T.Calls do
         seconds = call.ended_at |> DateTime.diff(call.accepted_at)
         minutes = div(seconds, 60)
 
+        if minutes > 5 do
+          T.Matches.match_after_live_call([call.caller_id, call.called_id])
```
ruslandoga, line 217: @akhrail1996 is this enough to decide if the call has been successful?

lib/t/matches.ex
```diff
@@ -265,6 +265,10 @@ defmodule T.Matches do
     |> Repo.one()
   end
 
+  def match_after_live_call(user_ids) do
+    [user_id_1, user_id_2] = Enum.sort(user_ids)
```
ruslandoga, line 269: @akhrail1996 what side-effects should there be, notifications, etc?

lib/t/calls.ex
akhrail1996: I guess even 1 minute call is enough

lib/t/matches.ex
akhrail1996: notifications are not needed, maybe just a broadcast so that LIVE Feed updates its UI (turning mate from regular profile into a match) - but I would leave that for later

as for Normal Feed no side-effects are needed, since both users will join Normal Feed freshly (therefore they will be rendered as matches correctly already at this point)

ruslandoga: moved to https://github.com/getsince/test3/issues/476

akhrail1996 (approved):

akhrail1996 (approved):

PR #470, opened 2021-12-26
track users in calls with phx tracker

ruslandoga:
This PR adds ability to track users who are in a call.


lib/t/calls.ex
```diff
@@ -35,19 +38,21 @@ defmodule T.Calls do
...
     end
   end
 
-  def call_allowed?(caller_id, called_id, reference) do
-    missed?(called_id, caller_id) or matched?(caller_id, called_id) or
-      in_live_mode?(caller_id, called_id, reference)
+  @spec ensure_call_allowed(uuid, uuid, DateTime.t()) :: :ok | {:error, reason :: String.t()}
+  defp ensure_call_allowed(caller_id, called_id, reference) do
+    cond do
+      TWeb.CallTracker.in_call?(called_id) -> {:error, dgettext("calls", "receiver is busy")}
```
akhrail1996, line 47: maybe it makes sense to localise error text on frontend?

akhrail1996 (approved):

lib/t/calls.ex
ruslandoga: I'll remove the gettext macros.

PR #475, opened 2021-12-29
Speed up tests, and make sure tests work offline

ruslandoga:
I ran `mix test --slowest 5` and noticed that these tests slowest either waited for network (apple sign in keys http request) or just waited (`:timer.sleep` in `update_last_active` test).

I replaced HTTP requests with a stub which returns current apple sign in keys, and used dependency injection for time in the  `update_last_active`.

I also disabled `Finch` in tests to reduce noise from its errors when it can't connect to apple servers when I'm offline.


akhrail1996 (approved):
👍

PR #477, opened 2021-12-29
allow audio uploads

ruslandoga:


lib/t/voicemail.ex
```diff
@@ -0,0 +1,18 @@
...
+  def audio_s3_url do
+    Media.user_s3_url()
+  end
+
+  def audio_upload_form(content_type) do
+    Media.sign_form_upload(
+      key: Ecto.UUID.generate(),
+      content_type: content_type,
+      # 50 MB
+      max_file_size: 50_000_000,
```
ruslandoga, line 14: or we can make the client provide us the size it want to send

lib/t_web/controllers/media_controller.ex
```diff
@@ -1,18 +1,26 @@
...
+  # if it's an image -> it's for profile story photo
+  defp upload_form("image/" <> _rest = content_type) do
     {:ok, %{"key" => key} = fields} = Accounts.photo_upload_form(content_type)
-    url = Accounts.photo_s3_url()
+    %{url: Accounts.photo_s3_url(), key: key, fields: fields}
+  end
 
-    json(conn, %{url: url, key: key, fields: fields})
+  # if it's an audio/aac -> it's for voicemail
+  defp upload_form("audio/aac" = content_type) do
```
akhrail1996, line 22: any rationale for choosing `aac` specifically?

akhrail1996 (approved):
thanks 🙏

lib/t_web/controllers/media_controller.ex
ruslandoga: does apple support anything else?

lib/t/voicemail.ex
akhrail1996: I will experiment with recordings on iOS to see the file sizes, but this should probably suffice for now

lib/t_web/controllers/media_controller.ex
akhrail1996: https://stackoverflow.com/questions/6867994/what-are-the-formats-supported-in-avaudiorecorder-for-recording-sound
seems to support some, yeah

ruslandoga: for recording ...

PR #481, opened 2021-12-29
fix n+1 db calls in views

### LINKED ISSUE #480: remove db calls from views
https://github.com/getsince/test3/blob/c938e8ae9200cac79c518822e310e92411379ab7/lib/t_web/views/match_view.ex#L10

ruslandoga:
closes #480

### DIFF
--- a/lib/t/matches.ex
+++ b/lib/t/matches.ex
@@ -73,19 +73,31 @@ defmodule T.Matches do
   end
 
   defp maybe_notify_match(
-         %Match{id: match_id},
+         %Match{id: match_id} = match,
          [by_user_id_settings, user_id_settings],
          by_user_id,
          user_id
        ) do
     broadcast_from_for_user(
       by_user_id,
-      {__MODULE__, :matched, %{id: match_id, mate: user_id, audio_only: user_id_settings}}
+      {__MODULE__, :matched,
+       %{
+         id: match_id,
+         expiration_date: expiration_date(match),
+         mate: user_id,
+         audio_only: user_id_settings
+       }}
     )
 
     broadcast_for_user(
       user_id,
-      {__MODULE__, :matched, %{id: match_id, mate: by_user_id, audio_only: by_user_id_settings}}
+      {__MODULE__, :matched,
+       %{
+         id: match_id,
+         expiration_date: expiration_date(match),
+         mate: by_user_id,
+         audio_only: by_user_id_settings
+       }}
     )
   end
 
@@ -302,18 +314,35 @@ defmodule T.Matches do
 
   @spec list_matches(uuid) :: [%Match{}]
   def list_matches(user_id) do
-    Match
-    |> where([m], m.user_id_1 == ^user_id or m.user_id_2 == ^user_id)
-    |> where([m], m.id not in subquery(archived_match_ids_q(user_id)))
+    matches_with_undying_events_q()
+    |> where([match: m], m.user_id_1 == ^user_id or m.user_id_2 == ^user_id)
+    |> where([match: m], m.id not in subquery(archived_match_ids_q(user_id)))
     |> order_by(desc: :inserted_at)
-    |> join(:left, [m], t in Timeslot, on: m.id == t.match_id)
-    |> join(:left, [m, t], c in MatchContact, on: m.id == c.match_id)
-    |> select([m, t, c], {m, t, c})
+    |> join(:left, [match: m], t in Timeslot, as: :t, on: m.id == t.match_id)
+    |> join(:left, [match: m], c in MatchContact, as: :c, on: m.id == c.match_id)
+    |> select([match: m, undying_event: e, t: t, c: c], {m, t, c, e.timestamp})
     |> Repo.all()
-    |> Enum.map(fn {match, timeslot, contact} ->
-      %Match{match | timeslot: timeslot, contact: contact}
+    |> Enum.map(fn {match, timeslot, contact, undying_event_timestamp} ->
+      expiration_date =
+        unless undying_event_timestamp do
+          expiration_date(match)
+        end
+
+      %Match{
+        match
+        | timeslot: timeslot,
+          contact: contact,
+          expiration_date: expiration_date,
+          interaction: timeslot || contact
+      }
     end)
     |> preload_match_profiles(user_id)
+  end
+
+  defp expiration_date(%Match{inserted_at: inserted_at}) do
+    inserted_at
+    |> DateTime.from_naive!("Etc/UTC")
+    |> DateTime.add(@match_ttl)
   end
 
   defp archived_match_ids_q(user_id) do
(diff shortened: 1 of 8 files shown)


lib/t/matches.ex
```diff
@@ -1055,23 +1062,27 @@ defmodule T.Matches do
...
+  def match_soon_to_expire_check(reference \\ DateTime.utc_now()) do
+    day_before_expiration = reference |> DateTime.add(-@match_ttl) |> DateTime.add(24 * 3600)
+    day_and_one_minute_before_expiration = DateTime.add(day_before_expiration, 60)
+
     expiring_matches_q()
-    |> where([m, e, c], e.timestamp <= fragment("now() - interval '6 days'"))
-    |> where([m, e, c], e.timestamp > fragment("now() - interval '6 days 1 minute'"))
-    |> select([m, e, c], m.id)
+    |> where([m], m.inserted_at <= ^day_before_expiration)
+    # TODO can result in duplicates with more than one worker
```
ruslandoga, line 1083: - [ ] create issue

test/t_web/channels/feed_channel_test.exs
```diff
@@ -72,17 +72,18 @@ defmodule TWeb.FeedChannelTest do
...
                %{
                  "id" => m2.id,
                  "profile" => %{name: "mate-2", story: [], user_id: p2.id, gender: "N"},
-                 "contact" => %{
-                   "contacts" => %{"whatsapp" => "+79666666666"},
-                   "picker" => p2.id,
-                   "opened_contact_type" => nil
+                 "timeslot" => %{
+                   "picker" => me.id,
+                   "slots" => slots
```
ruslandoga, line 83: @akhrail1996 it appears that this test was wrong, according to [the comment above](https://github.com/getsince/test3/blob/c938e8ae9200cac79c518822e310e92411379ab7/test/t_web/channels/feed_channel_test.exs#L60)

test/t_web/channels/feed_channel_test.exs
```diff
@@ -72,17 +72,18 @@ defmodule TWeb.FeedChannelTest do
                  "id" => m3.id,
                  "profile" => %{name: "mate-3", story: [], user_id: p3.id, gender: "M"},
                  "timeslot" => %{"selected_slot" => s2},
-                 "audio_only" => false
+                 "audio_only" => false,
+                 "expiration_date" => ~U[2021-10-07 12:16:07Z]
```
ruslandoga, line 76: @akhrail1996 since the matches didn't have started calls and reported meetings, `expiration_date` should be returned

lib/t/matches.ex
```diff
@@ -302,20 +314,37 @@ defmodule T.Matches do
...
+    |> Enum.map(fn {match, timeslot, contact, undying_event_timestamp} ->
+      expiration_date =
+        unless undying_event_timestamp do
+          expiration_date(match)
+        end
+
+      %Match{
+        match
+        | timeslot: timeslot,
+          contact: contact,
```
akhrail1996, line 334: does %Match{} need `timeslot` & `contact` when it has interaction?

ruslandoga:
@akhrail1996 I didn't understand why some of the matches without calls and meeting reports didn't include `expiration_date` in feed channel payloads, was it a mistake or by design?

lib/t/matches.ex
ruslandoga: moved to https://github.com/getsince/test3/issues/482

test/t_web/channels/feed_channel_test.exs
```diff
@@ -72,17 +72,18 @@ defmodule TWeb.FeedChannelTest do
...
                },
                %{
                  "id" => m2.id,
                  "profile" => %{name: "mate-2", story: [], user_id: p2.id, gender: "N"},
-                 "contact" => %{
-                   "contacts" => %{"whatsapp" => "+79666666666"},
-                   "picker" => p2.id,
-                   "opened_contact_type" => nil
+                 "timeslot" => %{
```
akhrail1996, line 81: > # add contact to m2, but it won't be returned since there is a timeslot as well
(from https://github.com/getsince/test3/pull/481/files#diff-ecf0850f17c63f59b2690d88d79fbd691819ed852ee56937771756ddd24c19c8L60) is incorrect, I undestand now why you were confused  - I forgot to delete this comment when making some of the changes (sorry for this 🙏)

so actually "whatsapp" contact should be returned here

akhrail1996: since contact was shared later than timeslot -> was supposed to overwrite it

akhrail1996 (changes requested):

test/t_web/channels/feed_channel_test.exs
akhrail1996: @ruslandoga actually it's the other way around - test is correct, but comment was wrong 🙃

lib/t/matches.ex
ruslandoga: why not?

test/t_web/channels/feed_channel_test.exs
akhrail1996: @ruslandoga yes, I don't remember why exactly it wasn't rendered here (though it didn't affect real usage) - but it should be

akhrail1996:
> @akhrail1996 I didn't understand why some of the matches without calls and meeting reports didn't include `expiration_date` in feed channel payloads, was it a mistake or by design?

@ruslandoga yes, it was a mistake

lib/t/matches.ex
akhrail1996: no reason
rather wanted to understand rationale - maybe you had something in mind where they will be needed later or something.

test/t_web/channels/feed_channel_test.exs
ruslandoga: I've updated the test and it's still wrong, the tests was successful before because of the order function clauses matched in the view.

lib/t/matches.ex
```diff
@@ -302,20 +314,37 @@ defmodule T.Matches do
...
+          expiration_date: expiration_date,
+          interaction: timeslot || contact
```
ruslandoga, line 336: @akhrail1996 the previous behaviour int he match_view is `contact || timeslot`

test/t_web/channels/feed_channel_test.exs
akhrail1996: for some reason I thought that in this PR `interaction` is supposed (?) to determine the latest interaction mode => here the latest interaction mode would be `contact` then

nevertheless the current PR is probably safe to merge as well, since current client wouldn't allow for `timeslot` and `contact` to co-exist anyways (user would need to "cancel" contact / timeslot to send timeslot / contact)

so we can implement the above functionality (`interaction` being based on the latest interaction mode or rather one new interaction mode overriding content of the previous one -> guarantee of one interaction mode at a time for match) in a separate PR (probably in voicemail one)

akhrail1996: > since current client wouldn't allow for timeslot and contact to co-exist anyways (user would need to "cancel" contact / timeslot to send timeslot / contact)

but in the next version of the client I plan to remove this from client -> needs to be done in `voicemail` PR

akhrail1996 (approved):

test/t_web/channels/feed_channel_test.exs
ruslandoga: > for some reason I thought that in this PR interaction is supposed

no,  this pr refactors match_view db calls which were to calculate expiration

ruslandoga: no, I don't think it makes sense to do in voicemail PR

ruslandoga: moved to https://github.com/getsince/test3/issues/483

PR #490, opened 2021-12-31
update live session & match expiration / insertion dates feed_channel returns

akhrail1996:
To support changes here: https://github.com/getsince/test5/pull/196/commits/1dc8ac675443ceed974c2413b965cbeab448f6bc

### DIFF
--- a/lib/t_web/channels/feed_channel.ex
+++ b/lib/t_web/channels/feed_channel.ex
@@ -47,7 +47,6 @@ defmodule TWeb.FeedChannel do
     Feeds.maybe_activate_session(user_id, gender, feed_filter)
 
     {_type, [session_start, session_end]} = Feeds.live_today(utc_now(socket))
-    live_session_duration = DateTime.diff(session_end, session_start)
 
     missed_calls =
       user_id
@@ -66,8 +65,8 @@ defmodule TWeb.FeedChannel do
 
     reply =
       %{"mode" => "live"}
-      |> Map.put("session_expiration_date", session_end)
-      |> Map.put("live_session_duration", live_session_duration)
+      |> Map.put("session_start_date", session_start)
+      |> Map.put("session_end_date", session_end)
       |> maybe_put("missed_calls", missed_calls)
       |> maybe_put("invites", invites)
       |> maybe_put("matches", matches)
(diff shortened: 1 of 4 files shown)


lib/t_web/channels/feed_channel.ex
```diff
@@ -66,8 +65,8 @@ defmodule TWeb.FeedChannel do
 
     reply =
       %{"mode" => "live"}
-      |> Map.put("session_expiration_date", session_end)
-      |> Map.put("live_session_duration", live_session_duration)
+      |> Map.put("session_start_date", session_start)
```
ruslandoga, line 68: @akhrail1996 I'd keep it, otherwise it's a breaking change for the clients.

lib/t_web/channels/feed_channel.ex
```diff
@@ -572,13 +570,15 @@ defmodule TWeb.FeedChannel do
...
         id: match_id,
+        inserted_at: inserted_at,
         audio_only: audio_only,
         profile: profile,
         interaction: interaction,
         expiration_date: expiration_date
       } ->
         render_match(%{
           id: match_id,
+          inserted_at: DateTime.from_naive!(inserted_at, "Etc/UTC"),
```
ruslandoga, line 581: I'd pass `match.inserted_at` as is, and ensure it's UTC in the view, that way we cover more area.

ruslandoga: https://github.com/getsince/test3/pull/492/files

lib/t_web/channels/feed_channel.ex
akhrail1996: @ruslandoga we can keep them but I planned to force-upgrade for clients anyways (since old clients do not support `voicemail` which is crucial now) -> doesn't make much sense (unless we want to release this backend prior to app release - which may be beneficial, but probably requires extra work)

for App Store review I made the app support current backend as well

ruslandoga: > but probably requires extra work

At least on the backend, there is no other breaking change.

akhrail1996: @ruslandoga yeah okay, I will put it back then (a bit of extra work is putting it back & forth into tests 😉)

ruslandoga:
@akhrail1996 I've rebased this PR against current `add-voice-mail`

PR #494, opened 2021-12-31
add voicemail timer

ruslandoga:


lib/t/matches.ex
```diff
@@ -1183,19 +1189,27 @@ defmodule T.Matches do
...
+  # 0 -> nobody
+  # 1 -> only one user in a match
+  # 2 -> both users exchanged voicemail
+  defp matches_with_voicemail_participants_count_q(query \\ named_match_q()) do
+    voicemail_count_q =
+      Voicemail
+      |> group_by([v], v.match_id)
+      |> select([v], count(v.caller_id, :distinct))
+
+    join(query, :left, [m], v in subquery(voicemail_count_q), on: v.match_id == m.id)
```
ruslandoga, line 1212: @akhrail1996 I think I'll use this for now instead of match_events for voicemail timer. I'll join on both "undying_events" and this count and it should provide enough info to find matches to expire:

1. has undying event -> no expire
2. no undying event and has callers count = 2 -> expire if match's seven days old(er)
3. no undying event and callers count < 2 -> expire if match's two days old(er)

ruslandoga: no longer relevant

PR #495, opened 2022-01-01
add match.exchanged_voicemail

ruslandoga:


lib/t/calls.ex
```diff
@@ -317,17 +317,49 @@ defmodule T.Calls do
...
+        [mate] = users -- [caller_id]
+        {:ok, mate}
+      else
+        {:error, :not_found}
+      end
+    end)
+    |> Multi.delete_all(:contacts, where(MatchContact, match_id: ^match_id))
+    |> Multi.delete_all(:timeslots, where(Timeslot, match_id: ^match_id))
+    |> Multi.insert(:voicemail, voicemail)
+    # TODO test
```
ruslandoga, line 341: - [ ] test

PR #491, opened 2021-12-31
add push and broadcast for voicemail

akhrail1996:

PR #497, opened 2022-01-01
add push and broadcast for voicemail 2

akhrail1996:
replacing https://github.com/getsince/test3/pull/491

### DIFF
--- a/lib/t_web/channels/feed_channel.ex
+++ b/lib/t_web/channels/feed_channel.ex
@@ -517,6 +517,17 @@ defmodule TWeb.FeedChannel do
     {:noreply, socket}
   end
 
+  def handle_info({Feeds, [:voicemail, :received], voicemail}, socket) do
+    %Calls.Voicemail{s3_key: s3_key, match_id: match_id} = voicemail
+
+    push(socket, "voicemail_received", %{
+      "match_id" => match_id,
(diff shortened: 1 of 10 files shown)


lib/t_web/channels/feed_channel.ex
```diff
@@ -517,6 +517,16 @@ defmodule TWeb.FeedChannel do
     {:noreply, socket}
   end
 
+  def handle_info({Feeds, [:voicemail, :sent], voicemail}, socket) do
```
ruslandoga, line 520:
```suggestion
  def handle_info({Feeds, [:voicemail, :received], voicemail}, socket) do
```

test/t/calls/voicemail_test.exs
```diff
@@ -0,0 +1,101 @@
+defmodule T.Calls.VoicemailTest do
+  use T.DataCase
```
ruslandoga, line 2: can be async

test/t/calls/voicemail_test.exs
```diff
@@ -0,0 +1,101 @@
+defmodule T.Calls.VoicemailTest do
```
ruslandoga, line 1: it might make sense to either move voicemail test cases from `T.CallsTest` or to move these tests there to stay consistent.

ruslandoga (approved):

PR #506, opened 2022-01-11
add accepted at to timeslots

ruslandoga:
Since the only place where `updated_at` could be used is `accept_slot`, I named it `accepted_at`.

The interactions can then be sorted like:

```elixir
interactions
|> Enum.sort_by(fn
    %Timeslot{inserted_at: inserted_at, accepted_at: accepted_at} -> accepted_at || inserted_at
    # other interactions
  end,
  {DateTime, :desc}
)
```

### DIFF
--- a/lib/t/matches.ex
+++ b/lib/t/matches.ex
@@ -710,6 +710,7 @@ defmodule T.Matches do
     )
 
     {:ok, slot, 0} = DateTime.from_iso8601(slot)
+    accepted_at = DateTime.truncate(reference, :second)
 
     {picker_name, _number_of_matches1} = user_info(picker)
     {mate_name, _umber_of_matches2} = user_info(mate)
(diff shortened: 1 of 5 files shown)


priv/repo/migrations/20220111113337_add_timeslots_updated_at.exs
```diff
@@ -0,0 +1,18 @@
+defmodule T.Repo.Migrations.AddTimeslotsAcceptedAt do
```
akhrail1996, line 1: filename mentions "updated_at" vs. "accepted_at" here

akhrail1996:
> Since the only place where `updated_at` could be used is `accept_slot`, I named it `accepted_at`.

A user may send one `timeslot_offer` and then alter it resulting -> `updated_at` needs to be updated

ruslandoga:
@akhrail1996 what function is called for altering?

akhrail1996:
@ruslandoga https://github.com/getsince/test3/blob/master/lib/t/matches.ex#L622

akhrail1996:
ah, my bad, this replaces_all on conflict

ruslandoga:
It would rewrite the prev timeslots for the match, and result in a new `inserted_at`.

akhrail1996 (approved):

PR #514, opened 2022-01-12
add match_contact.seen_at

ruslandoga:


lib/t/matches.ex
```diff
@@ -966,16 +966,17 @@ defmodule T.Matches do
     :ok
   end
 
-  def open_contact_for_match(me, match_id, contact_type) do
+  def open_contact_for_match(me, match_id, contact_type, now \\ DateTime.utc_now()) do
     m = "contact opened for match #{match_id} by #{me}"
+    seen_at = DateTime.truncate(now, :second)
```
ruslandoga, line 971: I thought that maybe at some point we'd want to know when it has been seen.

lib/t_web/channels/feed_channel.ex
```diff
@@ -326,7 +326,7 @@ defmodule TWeb.FeedChannel do
   def handle_in("open-contact", %{"match_id" => match_id, "contact_type" => contact_type}, socket) do
     me = me_id(socket)
 
-    :ok = Matches.open_contact_for_match(me, match_id, contact_type)
+    :ok = Matches.open_contact_for_match(me, match_id, contact_type, utc_now(socket))
```
ruslandoga, line 329: `utc_now` allows freezing time in tests

ruslandoga:
test failed but locally it works for the same seed: `mix test --seed 351287`, not sure what's wrong.

akhrail1996 (approved):

PR #517, opened 2022-01-13
fix timeslot start / end -> started / ended

akhrail1996:


akhrail1996:
@ruslandoga appreciate you leading me into order ;)

PR #521, opened 2022-01-15
dust off benches

ruslandoga:
I plan to add more benches so I dusted off the ones we currently have

### DIFF
--- a/bench/pubsub.exs
+++ b/bench/pubsub.exs
@@ -23,9 +23,10 @@ Enum.each(1..10, fn _ -> Sub.start_link(topic: "topic2") end)
 
 map = %{"some" => "key", "then" => "some oter key"}
 
-Benchee.run(%{
-  "topic_lots:message" => fn -> Phoenix.PubSub.broadcast(T.PubSub, "topic", "message") end,
-  "topic_lots:map" => fn -> Phoenix.PubSub.broadcast(T.PubSub, "topic", map) end,
-  "topic_few:message" => fn -> Phoenix.PubSub.broadcast(T.PubSub, "topic2", "message") end,
-  "topic_few:map" => fn -> Phoenix.PubSub.broadcast(T.PubSub, "topic2", map) end
-})
+Benchee.run(
+  %{
+    "message" => fn topic -> Phoenix.PubSub.broadcast(T.PubSub, topic, "message") end,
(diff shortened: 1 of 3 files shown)


bench/pubsub.exs
```diff
@@ -23,9 +23,10 @@ Enum.each(1..10, fn _ -> Sub.start_link(topic: "topic2") end)
...
-  "topic_few:message" => fn -> Phoenix.PubSub.broadcast(T.PubSub, "topic2", "message") end,
-  "topic_few:map" => fn -> Phoenix.PubSub.broadcast(T.PubSub, "topic2", map) end
-})
+Benchee.run(
+  %{
+    "message" => fn topic -> Phoenix.PubSub.broadcast(T.PubSub, topic, "message") end,
+    "map" => fn topic -> Phoenix.PubSub.broadcast(T.PubSub, topic, map) end
+  },
+  inputs: %{"lots subs" => "topic", "few subs" => "topic2"}
+)
```
ruslandoga, line 32: this is not really a fix but rather a small improvement

config/bench.exs
```diff
@@ -4,3 +4,4 @@ config :t, T.Media.Client, adapter: T.Media.S3Client
 config :t, T.Bot, adapter: T.Bot.API
 config :t, T.PushNotifications.APNS, adapter: T.PushNotifications.APNS.FinchAdapter
 config :t, T.Twilio, adapter: T.Twilio.HTTP
+config :t, T.Accounts.AppleSignIn, adapter: T.Accounts.AppleSignIn.HTTPAdapter
```
ruslandoga, line 7: this fixes benches compilation, it was failing without apple sign in adapter set since it's required during compilation (but it's not actually used in benches)

config/runtime.exs
```diff
@@ -301,6 +301,11 @@ if config_env() == :bench do
 
   config :t, T.Media.Static, disabled?: true
   config :t, Oban, queues: false, plugins: false
+  config :t, T.Feeds.SeenPruner, disabled?: true
+  config :t, T.Matches.MatchExpirer, disabled?: true
+  config :t, T.PushNotifications.ScheduledPushes, disabled?: true
+  config :t, T.Matches.TimeslotPruner, disabled?: true
+  config :t, Finch, disabled?: true
```
ruslandoga, line 308: I disabled all background tasks so as not to interfere with benches during their runs

akhrail1996 (approved):

PR #479, opened 2021-12-29
add voicemail

ruslandoga:
closes #474 
depends #494 #493 #491 #490 #484 #481 

- [x] merge all PRs
- [x] finish tests
- [x] https://github.com/getsince/test5/pull/196/files#diff-d5a373f3ff78460009f5de32b1aa475744f13f1125990d4de348896dc7c3bc42R57
- [x] don't delete prev interactions https://github.com/getsince/test3/pull/508
- [x] render all interactions per match https://github.com/getsince/test3/pull/509
- [ ] ~~sort interactions per match~~ can't be done right now since interactions are split between `contact`, `timeslot`, and `voicemail` fields https://github.com/getsince/test3/issues/510

PR #528, opened 2022-01-22
remove forced voicemail exchange

akhrail1996:


lib/t/calls.ex
```diff
@@ -322,42 +322,33 @@ defmodule T.Calls do
     Multi.new()
     |> Multi.run(:match, fn _repo, _changes ->
```
ruslandoga, line 323: before I added the new timer logic, this step was called `:mate`, maybe it makes sense to name it that again?

lib/t/calls.ex
```diff
@@ -322,42 +322,33 @@ defmodule T.Calls do
...
       match =
-        Matches.matches_with_undying_events_q()
+        Match
         |> where(id: ^match_id)
         |> where([m], m.user_id_1 == ^caller_id or m.user_id_2 == ^caller_id)
-        |> select([m, undying_event: e], {m.inserted_at, [m.user_id_1, m.user_id_2], e.timestamp})
+        |> select([m], {[m.user_id_1, m.user_id_2]})
```
ruslandoga, line 328:
```suggestion
        |> select([m], [m.user_id_1, m.user_id_2])
```

lib/t/calls.ex
```diff
@@ -322,42 +322,33 @@ defmodule T.Calls do
...
         |> Repo.one()
 
       case match do
-        {inserted_at, users, undying_event_at} ->
+        {users} ->
```
ruslandoga, line 332:
```suggestion
        users ->
```

ruslandoga: I think it might have been simpler before I added the timer logic, I'll check it out tomorrow.

lib/t/calls.ex
```diff
@@ -322,42 +322,33 @@ defmodule T.Calls do
...
-        |> Repo.update_all(set: [exchanged_voicemail: true])
 
-        {:ok, true}
-      else
-        {:ok, false}
-      end
+      {:ok, s3_keys}
     end)
     |> Multi.insert(:voicemail, voicemail)
     |> Oban.insert(:push, fn %{match: %{mate: mate}} ->
```
ruslandoga, line 354:
```suggestion
    |> Oban.insert(:push, fn :%{mate: mate} ->
```

lib/t/calls.ex
```diff
@@ -384,18 +374,9 @@ defmodule T.Calls do
...
-          unless undying_event_at do
-            if exchanged_voicemail do
-              inserted_at
-              |> DateTime.from_naive!("Etc/UTC")
-              |> DateTime.add(Matches.match_ttl())
-            end
-          end
-
-        broadcast_payload = %{voicemail: voicemail, expiration_date: new_expiration_date}
+        broadcast_payload = %{voicemail: voicemail}
```
ruslandoga, line 377:
```suggestion
        voicemail # no need to construct a new payload since there is no value other than voicemail in it
```

lib/t/matches.ex
```diff
@@ -326,29 +322,23 @@ defmodule T.Matches do
...
     |> Enum.map(fn {match, undying_event_timestamp} ->
       expiration_date =
-        cond do
-          undying_event_timestamp -> nil
-          match.exchanged_voicemail -> post_voicemail_expiration_date(match)
-          true -> pre_voicemail_expiration_date(match)
+        if undying_event_timestamp do
+          nil
+        else
+          expiration_date(match)
```
ruslandoga, line 328:
```suggestion
unless undying_event_timestamp do
  expiration_date(match)
end
```

lib/t/matches/match.ex
```diff
@@ -8,8 +8,6 @@ defmodule T.Matches.Match do
     field :user_id_1, Ecto.Bigflake.UUID
     field :user_id_2, Ecto.Bigflake.UUID
 
-    field :exchanged_voicemail, :boolean
```
ruslandoga, line 11: this seems like a breaking change, would old client be ok with this field removed from responses?

lib/t_web/channels/feed_channel.ex
```diff
@@ -424,8 +421,7 @@ defmodule TWeb.FeedChannel do
             profile: profile,
             screen_width: screen_width,
             inserted_at: inserted_at,
-            expiration_date: expiration_date,
-            exchanged_voice: false
```
ruslandoga, line 428: or rather this is could be the breaking change, since the client doesn't care about our DB but it does care about the immediate responses from backend

ruslandoga: is it ok to remove this two fields? how does the client handle their absence? or maybe it doesn't use them at all?

lib/t_web/channels/feed_channel.ex
```diff
@@ -517,7 +513,7 @@ defmodule TWeb.FeedChannel do
   end
 
   def handle_info({Calls, [:voicemail, :received], payload}, socket) do
-    %{voicemail: voicemail, expiration_date: expiration_date} = payload
+    %{voicemail: voicemail} = payload
```
ruslandoga, line 516:
```suggestion
   voicemail # if the suggestions from above are implemented
```

test/t/calls/voicemail_test.exs
```diff
@@ -31,32 +31,27 @@ defmodule T.Calls.VoicemailTest do
                Calls.voicemail_save_message(caller.id, match.id, s3_key)
     end
 
-    test "success: saves/overwrites voicemail, updates match.exchanged_voicemail" do
+    test "success: saves/overwrites voicemail" do
```
ruslandoga, line 34: this test and many others are redundant as they don't need to do the exchanges anymore to check for `exchanged_voicemail`

test/t/matches/expired_match_test.exs
```diff
@@ -22,50 +22,10 @@ defmodule T.Matches.ExpiredMatchTest do
...
   describe "expiration_list_soon_to_expire/0,1" do
-    defp in_pre_voicemail_exp_notification_window(match) do
-      match.inserted_at
-      |> DateTime.from_naive!("Etc/UTC")
-      |> DateTime.add(Matches.pre_voicemail_ttl())
-      |> DateTime.add(_3h = -3 * 3600)
-    end
-
-    defp in_post_voicemail_exp_notification_window(match) do
+    defp exp_notification_window(match) do
```
ruslandoga, line 28:
```suggestion
    defp in_exp_notification_window(match) do
```

ruslandoga: since the function returns a single value in the window, not the window itself, which would be a range

test/t_web/channels/feed_channel_test.exs
```diff
@@ -390,7 +379,6 @@ defmodule TWeb.FeedChannelTest do
       assert reply == %{
                "mode" => "normal",
                "since_live_date" => since_live_date,
-               "match_expiration_duration" => 172_800 = Matches.pre_voicemail_ttl(),
```
ruslandoga, line 393: just a note, I know this field needs to be removed but I believe it's best to make these changes in separate PRs otherwise it's easy to overlook ...

ruslandoga:
Would it be ok for me to push some commits on top in the morning? After that I think it would be ok to merge

lib/t/calls.ex
ruslandoga: https://github.com/getsince/test3/pull/479/commits/27d5d2ed0529d00ae0ec1615a82b3fe47a1b6554

lib/t/matches/match.ex
akhrail1996: with this in particular - yes (https://github.com/getsince/test5/pull/202/files#diff-d5a373f3ff78460009f5de32b1aa475744f13f1125990d4de348896dc7c3bc42L36);

with one other - no (https://github.com/getsince/test5/pull/202/files#diff-ff2ea48ff2f5f39ca8c03130d73e2dcb1e534ad449e4020fa77ec29657958c17L189)
-> I will revert this one so there is no breaking change

lib/t_web/channels/feed_channel.ex
akhrail1996: only `exchanged_voice` removed here, yeah it's okay

test/t/calls/voicemail_test.exs
akhrail1996: it checks saves and overwrites, ain't it?

akhrail1996:
@ruslandoga sure

PR #530, opened 2022-01-23
new socket

ruslandoga:


lib/t_web/socket/error_helpers.ex
```diff
@@ -0,0 +1,13 @@
+defmodule TWeb.Socket.ErrorHelpers do
+  @moduledoc false
+
+  errors = %{
+    call_not_allowed: [123, "call is not allowed"],
+    user_busy: [45, "user is busy"],
+    internal: [5000, "internal server error"]
```
ruslandoga, line 7: TODO need to be able to add changeset here as well

lib/t_web/socket/socket.ex
```diff
@@ -0,0 +1,103 @@
...
+  def init(req, [handler] = state) do
+    # TODO hibernate after 15 secs
+    # TODO check origin?
+    case handler.connect(req) do
+      {:ok, assigns} ->
+        {:cowboy_websocket, req, [handler | assigns], _opts = %{}}
+
+      {:error, status_code} ->
+        # TODO
+        {:ok, :cowboy_req.reply(status_code, req), state}
```
ruslandoga, line 32: this might not work with urlsessionwebsockettask since it doesn't return the status code in the error so we won't be able to distinguish between 403, 418, and actually invalid server response since all three get the same error in urlsessionwebsockettask

lib/t_web/socket/user_socket.ex
```diff
@@ -0,0 +1,270 @@
...
+      {:ok, call_id} ->
+        {:ok, %{"call_id" => call_id, "ice-servers" => [], "call_topics" => []}, assigns}
+
+      {:error, reason} ->
+        {:error, error(reason), assigns}
+    end
+  end
+
+  def handle_event("join-call", %{"user_id" => user_id}, assigns) do
+    {:ok, %{"caller" => %{}, "ice-servers" => [], "call_topics" => []}, assigns}
```
ruslandoga, line 214: actually we don't need join-call, we can return this info on `pick-up`

ruslandoga: this doesn't account for the case when `called` picks-up after `caller` hung-up, or rather it does but not in the best way:

```
1. caller calls
2. caller hangs up
3. called gets notification, connects to socket, picks up
4. pick up fails since the call has ended, called user is confused
```

ruslandoga: called user probably still needs to ask backend if the caller is still calling right after connecting to socket

```
1. caller calls
2. caller hangs up
3. called gets notification, connects to socket, pushes something like [1,"connect-call",{"call_id":"234"}]
4. receives `[1,"error",[404,"call has ended"]]`
5. this happens very quickly and the user doesn't notice there's been a call
```

ruslandoga: can be called `check-call`, `call-alive`, `connect-call`, `join-call` ...

PR #534, opened 2022-01-23
update ci to use new docker cache and github registry

### LINKED ISSUE #533: add arm64 images
[Graviton2 instances](https://aws.amazon.com/ec2/graviton/) seem nice.

And they'd be even nicer for us after [erlang 25](https://github.com/erlang/otp/pull/4869) is released in May with arm64 support for the [JIT](https://en.wikipedia.org/wiki/Just-in-time_compilation) compiler.

To be able to use them we'd need to [cross-compile the docker image to arm64.](https://github.com/getsince/test3/blob/e565b5ac123f50bf33a48ecc3043690dd652b4c4/.github/workflows/main.yml#L99-L145)

ruslandoga:
and closes #533 

Note that [hub.docker.com/ruslandoga/test3](https://hub.docker.com/repository/docker/ruslandoga/test3/general) would no longer be used, and [ghcr.io/getsince/test3](https://github.com/getsince/test3/pkgs/container/test3) would be used in its place. So [nomad](https://github.com/getsince/nomad/blob/0f1c43e10b8681603a4e06d33eef424420d6615f/backend2.nomad#L21) would need to use `ghcr.io/getsince/test3:<commit>` on the next deploy.


ruslandoga:
build took 23 minutes, but the next builds should take less since the deps are now cached

akhrail1996 (approved):

PR #540, opened 2022-01-31
add match interaction history

ruslandoga:
- [x] tests
- [x] broadcasts
- [x] add a function to populate `match_interactions` table: `Dev.populate_match_interactions(_dry_run = false)`
- [x] update function to populate `match_interactions` table

### DIFF
--- a/lib/t/calls.ex
+++ b/lib/t/calls.ex
@@ -8,7 +8,7 @@ defmodule T.Calls do
   alias T.{Repo, Twilio, Accounts, Feeds, Matches}
   alias T.Calls.{Call, Voicemail}
   alias T.Feeds.{FeedProfile}
-  alias T.Matches.{Match, MatchEvent, ArchivedMatch}
+  alias T.Matches.{Match, MatchEvent, ArchivedMatch, Interaction}
   alias T.PushNotifications.{APNS, DispatchJob}
   alias T.Bot
 
@@ -22,12 +22,31 @@ defmodule T.Calls do
   @spec call(uuid, uuid, DateTime.t()) :: {:ok, call_id :: uuid} | {:error, reason :: String.t()}
   def call(caller_id, called_id, reference \\ DateTime.utc_now()) do
     call_id = Ecto.Bigflake.UUID.generate()
+    call = %Call{id: call_id, caller_id: caller_id, called_id: called_id}
+
+    interaction =
+      if match_id = match_id(caller_id, called_id) do
+        %Interaction{
+          id: call_id,
+          match_id: match_id,
+          from_user_id: caller_id,
+          to_user_id: called_id,
+          data: %{"type" => "call_attempt"}
+        }
+      end
 
     with :ok <- ensure_call_allowed(caller_id, called_id, reference),
          {:ok, devices} <- ensure_has_devices(called_id),
          :ok <- push_call(caller_id, call_id, devices) do
-      %Call{id: ^call_id} =
-        Repo.insert!(%Call{id: call_id, caller_id: caller_id, called_id: called_id})
+      {:ok, maybe_interaction} =
+        Repo.transaction(fn ->
+          Repo.insert!(call)
+          if interaction, do: Repo.insert!(interaction)
+        end)
+
+      if maybe_interaction do
+        Matches.broadcast_interaction(maybe_interaction)
+      end
 
       m =
         "call attempt #{fetch_name(caller_id)} (#{caller_id}) calling #{fetch_name(called_id)} (#{called_id})"
@@ -67,6 +86,17 @@ defmodule T.Calls do
     |> where(user_id_1: ^user_id_1)
     |> where(user_id_2: ^user_id_2)
     |> Repo.exists?()
+  end
+
+  @spec match_id(uuid, uuid) :: uuid | nil
+  defp match_id(caller_id, called_id) do
+    [user_id_1, user_id_2] = Enum.sort([caller_id, called_id])
+
+    Match
+    |> where(user_id_1: ^user_id_1)
+    |> where(user_id_2: ^user_id_2)
+    |> select([m], m.id)
+    |> Repo.one()
   end
 
   defp reported?(caller_id, called_id) do
@@ -143,7 +173,7 @@ defmodule T.Calls do
     end)
   end
 
-  defp maybe_get_match_id(multi) do
+  defp get_match_id(multi) do
     Multi.run(multi, :match_id, fn _repo, %{call: {caller, called}} ->
       match_id =
         Match
@@ -158,14 +188,29 @@ defmodule T.Calls do
     end)
   end
 
-  defp maybe_insert_call_start_event(multi) do
+  defp maybe_insert_call_start_event(multi, now) do
     Multi.run(multi, :event, fn _repo, %{match_id: match_id} ->
       if match_id do
-        Repo.insert(%MatchEvent{
-          timestamp: DateTime.truncate(DateTime.utc_now(), :second),
+        event = %MatchEvent{timestamp: now, match_id: match_id, event: "call_start"}
+        Repo.insert(event)
+      else
+        {:ok, nil}
+      end
+    end)
+  end
+
+  defp maybe_insert_call_accepted_interaction(multi, call_id) do
+    Multi.run(multi, :interaction, fn _repo, %{match_id: match_id, call: {caller, called}} ->
+      if match_id do
+        interaction = %Interaction{
+          # id: call_id,
+          from_user_id: called,
+          to_user_id: caller,
           match_id: match_id,
-          event: "call_start"
-        })
+          data: %{"type" => "call_accepted", "call_id" => call_id}
+        }
+
+        Repo.insert(interaction)
       else
         {:ok, nil}
       end
@@ -177,20 +222,22 @@ defmodule T.Calls do
   def accept_call(call_id, now \\ utc_now()) do
     Multi.new()
     |> set_call_accepted_at(call_id, now)
-    |> maybe_get_match_id()
-    |> maybe_insert_call_start_event()
+    |> get_match_id()
+    |> maybe_insert_call_start_event(now)
+    # TODO overwrite prev interaction (call attempt)?
+    |> maybe_insert_call_accepted_interaction(call_id)
     |> Repo.transaction()
     |> case do
-      {:ok, %{call: {caller, called}, match_id: match_id}} = success ->
+      {:ok, %{call: {caller, called}, match_id: match_id, interaction: interaction}} = success ->
         if match_id do
           selected_slot =
-            T.Matches.Timeslot
+            Matches.Timeslot
             |> where(match_id: ^match_id)
             |> select([t], t.selected_slot)
             |> Repo.one()
 
           if selected_slot do
-            seconds = DateTime.utc_now() |> DateTime.diff(selected_slot)
+            seconds = DateTime.diff(now, selected_slot)
             minutes = div(seconds, 60)
 
             m =
@@ -200,7 +247,7 @@ defmodule T.Calls do
             Bot.async_post_message(m)
           end
 
-          T.Matches.notify_match_expiration_reset(match_id, [caller, called])
+          Matches.notify_match_expiration_reset(match_id, [caller, called])
         else
           m =
             "LIVE call starts #{fetch_name(caller)} (#{caller}) with #{fetch_name(called)} (#{called})"
@@ -209,6 +256,7 @@ defmodule T.Calls do
           Bot.async_post_message(m)
         end
 
+        Matches.broadcast_interaction(interaction)
         success
 
       {:error, _step, _reason, _changes} = failure ->
@@ -349,6 +397,15 @@ defmodule T.Calls do
       {:ok, s3_keys}
     end)
     |> Multi.insert(:voicemail, voicemail)
+    |> Multi.insert(:interaction, fn %{mate: mate, voicemail: voicemail} ->
+      %Interaction{
+        id: voicemail.id,
+        from_user_id: caller_id,
+        to_user_id: mate,
+        match_id: match_id,
+        data: %{"type" => "voicemail", "s3" => s3_key}
+      }
+    end)
     |> Oban.insert(:push, fn %{mate: mate} ->
       DispatchJob.new(%{
         "type" => "voicemail_sent",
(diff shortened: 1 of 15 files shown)


lib/t/calls.ex
```diff
@@ -22,12 +22,31 @@ defmodule T.Calls do
...
-      %Call{id: ^call_id} =
-        Repo.insert!(%Call{id: call_id, caller_id: caller_id, called_id: called_id})
+      {:ok, maybe_interaction} =
+        Repo.transaction(fn ->
+          Repo.insert!(call)
+          if interaction, do: Repo.insert!(interaction)
+        end)
+
+      if maybe_interaction do
+        Matches.broadcast_interaction(maybe_interaction)
```
ruslandoga, line 48: - [ ] test broadcast

lib/t/calls.ex
```diff
@@ -22,12 +22,31 @@ defmodule T.Calls do
...
 
     with :ok <- ensure_call_allowed(caller_id, called_id, reference),
          {:ok, devices} <- ensure_has_devices(called_id),
          :ok <- push_call(caller_id, call_id, devices) do
-      %Call{id: ^call_id} =
-        Repo.insert!(%Call{id: call_id, caller_id: caller_id, called_id: called_id})
+      {:ok, maybe_interaction} =
+        Repo.transaction(fn ->
+          Repo.insert!(call)
+          if interaction, do: Repo.insert!(interaction)
```
ruslandoga, line 44: - [ ] test saved

lib/t/calls.ex
```diff
@@ -209,6 +255,7 @@ defmodule T.Calls do
           Bot.async_post_message(m)
         end
 
+        Matches.broadcast_interaction(interaction)
```
ruslandoga, line 258: - [ ] test broadcast

lib/t/calls.ex
```diff
@@ -179,9 +209,25 @@ defmodule T.Calls do
     |> set_call_accepted_at(call_id, now)
     |> maybe_get_match_id()
     |> maybe_insert_call_start_event()
+    # TODO overwrite prev interaction (call attempt)?
+    |> Multi.run(:interaction, fn _repo, %{match_id: match_id, call: {caller, called}} ->
```
ruslandoga, line 213: - [ ] test saved

lib/t/calls.ex
```diff
@@ -360,14 +416,15 @@ defmodule T.Calls do
...
-      {:ok, %{voicemail: voicemail, mate: mate}} ->
+      {:ok, %{voicemail: voicemail, interaction: interaction, mate: mate}} ->
         m =
           "voicemail from #{fetch_name(caller_id)} (#{caller_id}) to #{fetch_name(mate)} (#{mate})"
 
         Logger.warn(m)
         Bot.async_post_message(m)
 
         Feeds.broadcast_for_user(mate, {__MODULE__, [:voicemail, :received], voicemail})
+        Matches.broadcast_interaction(interaction)
```
ruslandoga, line 427: - [ ] test broadcast

lib/t/matches.ex
```diff
@@ -695,17 +712,22 @@ defmodule T.Matches do
...
+      }
+    end)
     |> Oban.insert(:push, push_job)
     |> Repo.transaction()
     |> case do
-      {:ok, %{timeslot: %Timeslot{} = timeslot}} ->
+      {:ok, %{timeslot: %Timeslot{} = timeslot, interaction: interaction}} ->
         maybe_unarchive_match(match_id)
         broadcast_for_user(mate_id, {__MODULE__, [:timeslot, :offered], timeslot})
+        broadcast_interaction(interaction)
```
ruslandoga, line 730: - [ ] test

lib/t_web/channels/feed_channel.ex
```diff
@@ -387,6 +387,13 @@ defmodule TWeb.FeedChannel do
     {:reply, :ok, socket}
   end
 
+  # history
+
+  def handle_in("list-interactions", %{"match_id" => match_id}, socket) do
```
ruslandoga, line 392: - [ ] test

lib/t_web/channels/feed_channel.ex
```diff
@@ -577,6 +584,17 @@ defmodule TWeb.FeedChannel do
     {:noreply, socket}
   end
 
+  def handle_info({Matches, :interaction, interaction}, socket) do
+    %Matches.Interaction{match_id: match_id} = interaction
+
+    push(socket, "interaction", %{
```
ruslandoga, line 590: - [ ] test

lib/t/calls.ex
```diff
@@ -177,20 +222,22 @@ defmodule T.Calls do
   def accept_call(call_id, now \\ utc_now()) do
     Multi.new()
     |> set_call_accepted_at(call_id, now)
-    |> maybe_get_match_id()
-    |> maybe_insert_call_start_event()
+    |> get_match_id()
+    |> maybe_insert_call_start_event(now)
+    # TODO overwrite prev interaction (call attempt)?
```
akhrail1996, line 227: yes to this, and also `call_end` should probably overwrite all of these and include duration or (time start + time end) into `interaction` content

lib/t/matches.ex
```diff
@@ -934,22 +1008,41 @@ defmodule T.Matches do
     cancel_contacts(by_user_id, mate, match_id)
   end
 
+  # TODO remove
```
akhrail1996, line 1011: yes to this - can be safely removed

lib/t_web/views/match_view.ex
```diff
@@ -38,6 +38,74 @@ defmodule TWeb.MatchView do
...
+      "type" => type,
+      "slots" => slots,
+      "inserted_at" => datetime(id),
+      "picker" => picker
+    }
+  end
+
+  defp render_interaction("slot_accept" = type, interaction) do
+    %Interaction{id: id, data: %{"slot" => slot}} = interaction
+    %{"id" => id, "type" => type, "selected_slot" => slot, "accepted_at" => datetime(id)}
```
akhrail1996, line 60: `by_user_id` matters here for frontend

akhrail1996: also i would rename all timestamps for different interactions to `inserted_at`

lib/t/matches.ex
```diff
@@ -895,6 +958,19 @@ defmodule T.Matches do
...
       )
 
+    match_event = %MatchEvent{
+      timestamp: now,
+      match_id: match_id,
+      event: "contact_offer"
+    }
+
+    interaction = %Interaction{
+      data: %{"type" => "contact_offer", "contacts" => contacts},
```
akhrail1996, line 968: `contacts` should become just a single `contact` now (https://github.com/getsince/test3/pull/540/files#diff-431ecb538ba81baef40c63a16664a4e648660f51a14f8228b916130fe0ba2237R944)

I don't think it's needed, but if you want it to avoid breaking changes, we can still accept multiple contacts, but insert them one by one

lib/t_web/views/match_view.ex
```diff
@@ -38,6 +38,74 @@ defmodule TWeb.MatchView do
...
+  end
+
+  defp render_interaction("slot_cancel" = type, interaction) do
+    %Interaction{id: id, from_user_id: by_user_id} = interaction
+    %{"id" => id, "type" => type, "by_user_id" => by_user_id, "cancelled_at" => datetime(id)}
+  end
+
+  defp render_interaction("contact_offer" = type, interaction) do
+    %Interaction{id: id, data: %{"contacts" => contacts}, to_user_id: picker} = interaction
```
akhrail1996, line 69: accordingly this becomes a `contact`

lib/t_web/views/match_view.ex
akhrail1996: also I would rename all pickers / senders / callers into `by_user_id`

lib/t_web/views/match_view.ex
```diff
@@ -38,6 +38,74 @@ defmodule TWeb.MatchView do
...
+      "inserted_at" => datetime(id)
+    }
+  end
+
+  defp render_interaction("call_attempt" = type, interaction) do
+    %Interaction{id: id, from_user_id: caller} = interaction
+    %{"id" => id, "type" => type, "caller" => caller, "attempted_at" => datetime(id)}
+  end
+
+  defp render_interaction("call_accepted", interaction) do
```
akhrail1996, line 104: there also needs to be a `call_ended` interaction with duration

akhrail1996 (commented):
thanks 🙏

looking good, but I have to test it with frontend (hopefully tomorrow)

one more thing - I want to drop "meeting reported" thing as a timer expiration event => probably "contact opened" may fill its role instead (but maybe this is a separate PR into this one)

lib/t/matches.ex
ruslandoga: >  but if you want it to avoid breaking changes

What breaking change does using `contacts` instead of `contact` introduce?

akhrail1996: > What breaking change does using contacts instead of contact introduce?

I meant switching to single `contact` from current `contacts` will introduce a breaking change

ruslandoga:
> one more thing - I want to drop "meeting reported" thing as a timer expiration event => probably "contact opened" may fill its role instead (but maybe this is a separate PR into this one)

completely unrelated, it's a separate issue

lib/t/matches.ex
ruslandoga: I'm still using `contacts`, where is `contact` coming from?

akhrail1996:
> > one more thing - I want to drop "meeting reported" thing as a timer expiration event => probably "contact opened" may fill its role instead (but maybe this is a separate PR into this one)
> 
> completely unrelated, it's a separate issue
right: https://github.com/getsince/test3/issues/541

lib/t/matches.ex
akhrail1996: yes, I am suggesting switching to `contact` since `contacts` suited our previous model (we rendered all sent contacts at once in a popup), but do not suit our current model (where each contact is a separate chat item)

ruslandoga: what stops you from rendering contacts as separate chat items?

akhrail1996: order has to be preserved, i.e.:
i send you "telegram" contact
you send me voice message
i send you "whatsapp" contact

ruslandoga: order would be preserved

akhrail1996: the correct order will be:
1. "telegram" contact
2. voice message
3. "whatsapp" contact
I do not see how I will be able to preserve this order with current backend code, maybe I am missing something?

ruslandoga:
```swift
let interactions: [MatchInteraction] = [
  .contactOffer(contacts: Contacts(telegram: "...")),
  .voicemail(...),
  .contactOffer(contacts: Contacts(whatsapp: "...")),
]
```

akhrail1996: ah, right, I see where you are coming from now

but current frontend unfortunately doesn't work that way: https://github.com/getsince/test5/blob/rails-fork/Since/Feeds/Normal%20Feed/NormalFeedChannel.swift#L266
so it will effectively always append to contacts list making it:
1) "telegram" contact
2) voice message
3) "telegram" contact + "whatsapp" contact

akhrail1996: right, but if we enforce "one contact" at a time things should be named properly + for contact type `interaction` rendering frontend will need to pick just one contact out of these

ruslandoga: > we enforce "one contact" at a time

how?

akhrail1996: > How?

in current version of frontend there will be only one contact sent at a time

> No matter what approach you pick, you still can do it with both representations, since they are equivalent.

if there is only one interaction - i get exactly the first contact I am trying (e.g. if dict["telegram"] else if dict["whatsapp"] etc.); if there are many - I will get the first one I am asking for (which is probably still fine since the update curve is pretty steep + not that many users will send multiple contacts using old version)

ruslandoga:
```swift
struct Contact {
  let type: String
  let value: String
}

enum MatchInteraction {
  case contact(Contact)
  case voicemail
}

let interactions: [[String: Any]] = [
  ["type": "contact_offer", "contacts": ["telegram": "@telegram"]],
  ["type": "voicemail"],
  ["type": "contact_offer",  "contacts": ["whatsapp": "@whatsapp", "telegram": "@telegram"]]
]

func decode(_ json: [[String: Any]]) -> [MatchInteraction] {
  var seenContactsFromMe = MapSet<Contact>()
  json.compactMap { raw in
    guard let type = raw["type"] as? String else { return nil }
    
    switch type {
    case "contact_offer": 
    }
  }
}
```

lib/t_web/views/match_view.ex
ruslandoga: > by_user_id matters here for frontend

https://github.com/getsince/test3/pull/540/commits/74c5afbb460a2c419b4a059c186349f598dfe231

lib/t_web/views/match_view.ex
ruslandoga: Duration should be calculated on the client since backend doesn't know in what format / units the client wants it.

lib/t/calls.ex
ruslandoga: https://github.com/getsince/test3/pull/540/commits/433a640bda6f50af544e04314670d439223aa14b

lib/t_web/views/match_view.ex
ruslandoga: > also i would rename all timestamps for different interactions to inserted_at

https://github.com/getsince/test3/pull/540/commits/8b775fa2c7f30fc0aad1a61aa8738530e6dd4728

ruslandoga: > also I would rename all pickers / senders / callers into by_user_id

https://github.com/getsince/test3/pull/540/commits/0e39b08b61b130fa893092c2c40796c0fa063279

akhrail1996 (commented):
@ruslandoga thanks 🙏
can I start testing it with frontend?

ruslandoga:
yes, we can also deploy it if needed

ruslandoga:
if it's good enough, we can merge it as well

akhrail1996:
@ruslandoga 
okay, me close to finally testing it 😋

on a separate note - we really need to return at least something (id of last interaction / bool for presence of unreplied dateOfferes / unlistened voicemail / unopened contacts, etc.) with `Matches` in `feedChannel.join` - could you please add this to this PR?

ruslandoga:
I can add last interaction id to every match returned on join.

akhrail1996 (approved):
@ruslandoga flawless work 👏

let's merge & deploy & populate interactions

Issue #545, opened 2022-02-04
disable real "LIVE" mode


akhrail1996:
2 reasons:
- people are not using it much
- we don't have design (both product-wise & visual) for it yet (as part of chat redesign)

ruslandoga:
@akhrail1996 can live be removed?

akhrail1996:
@ruslandoga I still want to come up with the right approach to LIVE, but not sure when this will happen
so probably just disabling it the optimal for now

ruslandoga:
what should be done about the issues above?

akhrail1996:
all closed?

ruslandoga:
> Then https://github.com/getsince/test3/issues/467, https://github.com/getsince/test3/issues/505, and https://github.com/getsince/test3/issues/461 could be closed.

akhrail1996:
@ruslandoga, yes I am referring to them

ruslandoga:
So you want to just close them?

akhrail1996:
@ruslandoga okay then maybe it makes sense to say a deadline (say next weekend):
- if there will be some more material ideas (of how LIVE can work) + understanding of timing by then => keep alive
- no ideas => remove

sounds good?

akhrail1996:
@ruslandoga LIVE can be removed after all 😉

PR #548, opened 2022-02-05
disable Since LIVE

akhrail1996:
@ruslandoga I will deploy this so that:
a) users won't receive "there is Since LIVE today" push notification @ 1pm MSC
b) there won't be Since LIVE tomorrow (as potentially the new version will be out by then; and new version doesn't support Since LIVE)

~~but obviously we need to do this in a proper way, probably with is_live_enabled variable in .envrc file
could you please do it?~~

related #545


akhrail1996:
deployed

akhrail1996:
@ruslandoga thanks 🙏

ruslandoga (approved):

Issue #546, opened 2022-02-04
drop "meeting_reports" (from feed_channel's handle_in & everywhere else)


akhrail1996:

akhrail1996:
don't forget to delete this test:
https://github.com/getsince/test3/pull/549/commits/ad60b2f8e11b2fce13bfdd786aba6de1630a76c6

ruslandoga:
I think this issue can be closed since that functionality has been already dropped

PR #549, opened 2022-02-05
"meeting_report" => "contact_offer" as match undying event

### LINKED ISSUE #541: drop "meeting reported" as a timer expiration event with "contact send" filling its role instead

akhrail1996:
closes #541

### DIFF
--- a/lib/t/matches.ex
+++ b/lib/t/matches.ex
@@ -1254,7 +1254,7 @@ defmodule T.Matches do
     undying_events_q =
       MatchEvent
       |> where(match_id: parent_as(:match).id)
-      |> where([e], e.event == "call_start" or e.event == "meeting_report")
+      |> where([e], e.event == "call_start" or e.event == "contact_offer")
       |> select([e], e.timestamp)
       |> limit(1)
 
--- a/test/t_web/channels/feed_channel_test.exs
+++ b/test/t_web/channels/feed_channel_test.exs
@@ -1482,6 +1482,7 @@ defmodule TWeb.FeedChannelTest do
     end
   end
 
+  # test doesn't make sense since "report-we-met" doesn't affect match expiration now, should be removed
   describe "open-contact, report-we-met, report-we-not-met" do
     setup :joined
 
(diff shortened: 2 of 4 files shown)


lib/t/matches.ex
```diff
@@ -1254,7 +1254,7 @@ defmodule T.Matches do
     undying_events_q =
       MatchEvent
       |> where(match_id: parent_as(:match).id)
-      |> where([e], e.event == "call_start" or e.event == "meeting_report")
+      |> where([e], e.event == "call_start" or e.event == "contact_offer")
```
akhrail1996, line 1257: turned out we already had `contact_offer` match_event
and since `meeting_report` can only be inserted after `contact_offer` => no need to insert `contact_offer` for every `meeting_report` in a migration

test/t_web/channels/feed_channel_test.exs
```diff
@@ -1559,6 +1557,7 @@ defmodule TWeb.FeedChannelTest do
                  "id" => m1.id,
                  "profile" => %{name: "mate-1", story: [], user_id: p1.id, gender: "F"},
                  "audio_only" => false,
+                 "expiration_date" => ~U[2021-10-07 12:16:05Z],
```
akhrail1996, line 1560: this test doesn't make much sense, but will be removed in #546

ruslandoga: please add a short comment about why it doesn't make sense on top of the test and reference it in the issue

lib/t/matches.ex
ruslandoga: Right, I ran

```sql
select mr.match_id from (select * from match_events where event = 'meeting_report') mr
left join (select * from match_events where event = 'contact_offer') co on mr.match_id = co.match_id
where co.match_id is null;
```

and it returns 0 results. It checks that every meeting report has a corresponding contact offer

test/t_web/channels/feed_channel_test.exs
ruslandoga: thank you

akhrail1996: done

ruslandoga:
@akhrail1996 please try writing shorter branch names and longer commit messages

ruslandoga (approved):

PR #553, opened 2022-02-05
start oban before endpoint

ruslandoga:
fixes https://sentry.io/organizations/since-kindly/issues/2990267127/?project=5972006&query=is%3Aunresolved

### DIFF
--- a/lib/t/application.ex
+++ b/lib/t/application.ex
@@ -28,11 +28,11 @@ defmodule T.Application do
         TWeb.Presence,
         TWeb.UserSocket.Monitor,
         T.Repo,
+        maybe_migrator(),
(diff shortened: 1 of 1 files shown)


lib/t/application.ex
```diff
@@ -28,11 +28,11 @@ defmodule T.Application do
         TWeb.Presence,
         TWeb.UserSocket.Monitor,
         T.Repo,
+        maybe_migrator(),
```
ruslandoga, line 31: migrations need to run before oban since they create/update tables for it, although it doesn't matter anymore since the tables are already created

akhrail1996 (approved):

Issue #565, opened 2022-02-11
Add private stories


akhrail1996:
- [ ] add private_story and blurred_private_story to db
- [ ] save private stories in update_profile
- [ ] return blurred_private_story with normal feed profiles; private_story with match profiles
- [ ] return both private_story and blurred_private_story with own_profile

ruslandoga:
@akhrail1996 how do the payloads look like?

akhrail1996:
> @akhrail1996 how do the payloads look like?

Maybe:
- Update profile / own profile: ["private_story"] ["blurred_private_story]
- Feed profile: just append blurred-private_story to normal story?
- Match profile: ["private_story"] - we need to understand which stories are private

ruslandoga:
@akhrail1996 are these keys / array / events?

akhrail1996:
> @akhrail1996 are these keys / arrays / events? How do you send them right now?
> 
> ```swift
> let payload = // <-- what's in here? how does it look like when there is a private story?
> 
> channel.push("update-profile", payload: payload)
> ```

```swift
let patch = ["story": publicStory, "private_story": privateStory, "blurred_private_story": blurredPrivateStory]

channel.push("submit", payload: ["profile": patch], timeout: timeout)
```
https://github.com/getsince/test5/pull/209/commits/31ced8c2a6216e590217b36e3b403734a3b4d253#diff-fae0949f9c32a4dd3fb697f321e771901d0d32997336e519017ffda9f5283ff7R111

akhrail1996:
> * ~Feed profile: just append blurred-private_story to normal story?~

I think it's better & more correct for normal feed profiles to also have ["private_story"] containing  `blurred_private_story`

also on `"like"` resulting in a match "private_story" has to be returned here: https://github.com/getsince/test3/blob/master/lib/t_web/channels/feed_channel.ex#L156

ruslandoga:
@akhrail1996 what's in `privateStory` and `blurredPrivateStory`. Please give me the whole payload.

PR #571, opened 2022-02-14
add private stories

ruslandoga:
closes https://github.com/getsince/test3/issues/565

### DIFF
--- a/lib/t_web/views/feed_view.ex
+++ b/lib/t_web/views/feed_view.ex
@@ -4,25 +4,28 @@ defmodule TWeb.FeedView do
   alias T.Feeds.{FeedProfile}
 
   def render("feed_item.json", %{profile: profile, screen_width: screen_width}) do
-    %{"profile" => render_profile(profile, [:user_id, :name, :gender, :story], screen_width)}
+    profile =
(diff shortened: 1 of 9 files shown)


lib/t_web/views/feed_view.ex
```diff
@@ -4,25 +4,28 @@ defmodule TWeb.FeedView do
...
-    render_profile(profile, [:user_id, :name, :gender, :story], screen_width)
+    render_profile(profile, [:user_id, :name, :gender, :story], screen_width, _env = :match)
   end
 
   def render("missed_call.json", %{profile: profile, call: call, screen_width: screen_width}) do
-    %{
-      "profile" => render_profile(profile, [:user_id, :name, :gender, :story], screen_width),
-      "call" => CallView.render("call.json", call: call)
-    }
+    profile =
```
akhrail1996, line 18: missed_call.json (& missed calls) can be deleted, version of the app with chat (93% adoption) do not use it at all; and for older version this won't be a breaking change as well

akhrail1996 (approved):
seems okay, I will test before merging

akhrail1996 (approved):
works as expected. thanks!

PR #578, opened 2022-02-16
add 'deprecated' alerts'

ruslandoga:


ruslandoga:
I'll remove the attempt to allow users on old version to interact, add alerts on all deprecated messages, and localize

akhrail1996 (commented):
I just realised I fucked up when releasing minor patch with deprecated alerts support by adding it everywhere except it's needed the most 🙈
fixed in https://github.com/getsince/test5/pull/211/commits/95628270e9afaab1b1ab3ccb92d55119e43c3fa5
releasing it now

akhrail1996 (changes requested):
this works as expected, however I think we need to add `send-contact` here as well

priv/gettext/ru/LC_MESSAGES/alerts.po
```diff
@@ -0,0 +1,38 @@
...
+msgstr ""
+"Language: ru\n"
+"Plural-Forms: nplurals=3\n"
+
+#, elixir-autogen, elixir-format
+#: lib/t_web/channels/feed_channel.ex:172
+#: lib/t_web/channels/feed_channel.ex:235
+#: lib/t_web/channels/feed_channel.ex:245
+msgid "Calls are no longer supported, please upgrade."
+msgstr "Звонки больше не работают, пожалуйста обновитесь."
```
ruslandoga, line 19: В остальных сообщениях используется "ты" или "вы"?

akhrail1996: везде используется "ты"

akhrail1996: я исправлю

lib/t/accounts.ex
```diff
@@ -444,6 +444,20 @@ defmodule T.Accounts do
     get_user_by_session_token_and_update_version(token, nil, context)
   end
 
+  @spec list_app_versions(Ecto.Bigflake.UUID.t()) :: [String.t()]
+  def list_app_versions(user_id) do
+    UserToken |> where(user_id: ^user_id) |> select([t], t.version) |> Repo.all()
+  end
+
+  @spec has_previous_version?(Ecto.Bigflake.UUID.t()) :: bool
+  def has_previous_version?(user_id) do
```
akhrail1996, line 453: @ruslandoga is this and the command above used anywhere?
couldn't find. can I remove them?

ruslandoga: Not used.

I wanted to use it for stats, but now I see that we can just read it from DB. I'd be very thankful if you remove them. 

Thank you.

priv/gettext/ru/LC_MESSAGES/alerts.po
ruslandoga: спасибо

lib/t_web/channels/feed_channel.ex
```diff
@@ -266,7 +266,7 @@ defmodule TWeb.FeedChannel do
         dgettext("alerts", "Sending contacts is no longer supported, please upgrade.")
       )
 
-    {:reply, {:error, %{alerts: alert}}, socket}
+    {:reply, {:error, %{alert: alert}}, socket}
```
ruslandoga, line 266: 👍

akhrail1996 (approved):

akhrail1996:
will merge & deploy upon release

PR #593, opened 2022-02-20
add stickers, position, zoom, rotation on /admin/profile labels

ruslandoga:


lib/t/media/static.ex
```diff
@@ -57,7 +57,7 @@ defmodule T.Media.Static do
   def init(_opts) do
     @table = :ets.new(@table, [:named_table])
     Phoenix.PubSub.subscribe(@pubsub, @topic)
-    :timer.send_interval(:timer.minutes(1), :refresh)
+    # :timer.send_interval(:timer.minutes(1), :refresh)
```
ruslandoga, line 60: - [ ] reset

PR #599, opened 2022-02-22
add likes.seen

ruslandoga:


lib/t/matches.ex
```diff
@@ -236,6 +236,16 @@ defmodule T.Matches do
     Multi.insert(multi, :like, changeset)
   end
 
+  @spec mark_like_seen(uuid, uuid) :: :ok
```
ruslandoga, line 239: - [ ] test

lib/t_web/channels/feed_channel.ex
```diff
@@ -195,6 +195,11 @@ defmodule TWeb.FeedChannel do
     {:reply, reply, socket}
   end
 
+  def handle_in("seen-like", %{"user_id" => by_user_id}, socket) do
```
ruslandoga, line 198: - [ ] test

lib/t_web/channels/feed_channel.ex
```diff
@@ -475,6 +480,14 @@ defmodule TWeb.FeedChannel do
     Enum.map(feed, fn feed_item -> render_feed_item(feed_item, screen_width) end)
   end
 
+  defp render_likes(likes, screen_width) do
+    Enum.map(likes, fn %{profile: profile, seen: seen} ->
+      profile
+      |> render_feed_item(screen_width)
+      |> maybe_put("seen", seen)
```
ruslandoga, line 487: - [ ] test when seen=true

lib/t_web/channels/feed_channel.ex
akhrail1996, line 198: shouldn't we be recording seenStats timings here as well? similar to: https://github.com/getsince/test3/commit/372ba575981602ff588ee53828171274aabd43f2#diff-a0218a402f6110a957246806c0683922fda46cc95037890c7aae13528effb6a0R118

frontend is sending them: https://github.com/getsince/test5/pull/228/files#diff-525f3ab56ece6ac64f5c6cf950b0f62c83aefde5e7b176c28e2ff302ee9fc805R202

ruslandoga: done

Issue #602, opened 2022-02-22
maybe move existing contact labels & even ever shared contacts (?) to private pages (?)


akhrail1996:

akhrail1996:
@ruslandoga 
this can be used as a "placeholder" blurred page background for profiles with only email:

url: https://since-when-are-you-happy.s3.amazonaws.com/48f74064-ddc3-4072-89d7-cf0da9415d4e
s3key: 48f74064-ddc3-4072-89d7-cf0da9415d4e

ruslandoga:
@akhrail1996 this seems to be done

PR #605, opened 2022-02-24
add version-based story rendering

ruslandoga:
blocks https://github.com/getsince/test3/issues/602

### DIFF
--- a/lib/t_web/controllers/mobile_auth_controller.ex
+++ b/lib/t_web/controllers/mobile_auth_controller.ex
@@ -21,10 +21,13 @@ defmodule TWeb.MobileAuthController do
       token: Accounts.UserToken.encoded_token(token),
       user: render_user(user),
       # TODO proper screen_width
+      # TODO actually, should it be removed?
       profile:
         TWeb.ProfileView.render("show_with_location.json",
           profile: user.profile,
-          screen_width: 1000
+          screen_width: 1000,
+          # TODO update to 6.0.0
+          version: "5.2.3"
         )
     })
   end
--- a/lib/t_web/controllers/mobile_auth_controller.ex
+++ b/lib/t_web/controllers/mobile_auth_controller.ex
@@ -27,7 +27,7 @@ defmodule TWeb.MobileAuthController do
           profile: user.profile,
           screen_width: 1000,
           # TODO update to 6.0.0
-          version: "5.2.3"
+          version: "6.0.0"
         )
     })
   end
--- a/lib/t_web/controllers/mobile_auth_controller.ex
+++ b/lib/t_web/controllers/mobile_auth_controller.ex
@@ -20,13 +20,11 @@ defmodule TWeb.MobileAuthController do
     |> json(%{
       token: Accounts.UserToken.encoded_token(token),
       user: render_user(user),
-      # TODO proper screen_width
-      # TODO actually, should it be removed?
+      # TODO remove in `new-socket`
       profile:
         TWeb.ProfileView.render("show_with_location.json",
           profile: user.profile,
           screen_width: 1000,
-          # TODO update to 6.0.0
           version: "6.0.0"
         )
     })
--- a/lib/t_web/views/view_helpers.ex
+++ b/lib/t_web/views/view_helpers.ex
@@ -4,11 +4,32 @@ defmodule TWeb.ViewHelpers do
 
   @type env :: :feed | :match | :profile
 
-  @spec postprocess_story([map], pos_integer(), env) :: [map]
-  def postprocess_story(story, screen_width, env) when is_list(story) do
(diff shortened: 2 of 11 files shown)


lib/t_web/views/view_helpers.ex
```diff
@@ -4,13 +4,34 @@ defmodule TWeb.ViewHelpers do
 
   @type env :: :feed | :match | :profile
 
-  @spec postprocess_story([map], pos_integer(), env) :: [map]
-  def postprocess_story(story, screen_width, env) when is_list(story) do
+  @spec postprocess_story([map], String.t(), pos_integer(), env) :: [map]
+  def postprocess_story(story, "6." <> _rest, screen_width, env) when is_list(story) do
```
ruslandoga, line 8: main changes are in this module

lib/t_web/controllers/mobile_auth_controller.ex
```diff
@@ -21,10 +21,13 @@ defmodule TWeb.MobileAuthController do
       token: Accounts.UserToken.encoded_token(token),
       user: render_user(user),
       # TODO proper screen_width
+      # TODO actually, should it be removed?
```
ruslandoga, line 24: @akhrail1996 I don't think profile from HTTP response to the auth request it's being used in any version of the app, right?

lib/t_web/views/view_helpers.ex
```diff
@@ -38,24 +59,73 @@ defmodule TWeb.ViewHelpers do
...
-    end)
+      end)
+      |> :lists.reverse()
+
+    %{page | "labels" => labels}
+  end
+
+  defp process_labels_pre_v6(page), do: page
+
+  # TODO since handle is passed as in into the urls, it needs to be validated on insert
```
ruslandoga, line 102: - [ ] open issue

lib/t_web/controllers/mobile_auth_controller.ex
akhrail1996: @ruslandoga it's used for determining whether user is onboarded or not here: https://github.com/getsince/test5/blob/master/Since/Auth/AuthService.swift#L47

lib/t_web/controllers/mobile_auth_controller.ex
```diff
@@ -21,10 +21,13 @@ defmodule TWeb.MobileAuthController do
...
       profile:
         TWeb.ProfileView.render("show_with_location.json",
           profile: user.profile,
-          screen_width: 1000
+          screen_width: 1000,
+          # TODO update to 6.0.0
+          version: "5.2.3"
```
akhrail1996, line 30: that's why this should be indeed bumped to 6.0.0 (it's safe to decode 6.0.0 OwnProfile for 5.2.3 and below) because of an edge case where the only user's photo is in its private page (which won't be returned because of this)

lib/t_web/controllers/mobile_auth_controller.ex
akhrail1996: after that - no, since Current.ownProfile is taken from profileChannel: 
https://github.com/getsince/test5/blob/master/Since/Auth/AuthedCoordinator.swift#L71
https://github.com/getsince/test5/blob/master/Since/App/AppSocket.swift#L105

lib/t_web/controllers/mobile_auth_controller.ex
akhrail1996, line 30:
```suggestion
          version: "6.0.0"
```

akhrail1996 (changes requested):

lib/t_web/controllers/mobile_auth_controller.ex
ruslandoga: just in case, `user.onboarded_at` can be used to check if user has onboarded

lib/t_web/controllers/mobile_auth_controller.ex
ruslandoga: wouldn't ios crash on new stickers though?

akhrail1996: It's in 5.2.3:
https://github.com/getsince/test5/blob/add-seen-id-to-chat-version/Since/Story/Models/StoryLabel.swift#L106

lib/t_web/controllers/mobile_auth_controller.ex
akhrail1996: thanks, will explore that path

akhrail1996 (approved):

akhrail1996:
@ruslandoga do you think this versioning system will be useful going forward (for addition of new element types to stories)? or should it be removed after adoption of 6.0.0?

ruslandoga:
@akhrail1996 I don't know.

PR #606, opened 2022-02-25
add contact validation

ruslandoga:
adds basic contact validation on backend

### DIFF
--- a/lib/t/accounts/profile.ex
+++ b/lib/t/accounts/profile.ex
@@ -1,6 +1,7 @@
 defmodule T.Accounts.Profile do
   use Ecto.Schema
   import Ecto.Changeset
+  import T.Gettext
   alias T.Stickers
 
   @primary_key false
@@ -30,6 +31,12 @@ defmodule T.Accounts.Profile do
 
     # settings
     field :audio_only, :boolean, virtual: true
+  end
+
+  def changeset(profile, attrs, opts \\ []) do
+    profile
+    |> essential_info_changeset(attrs, opts)
+    |> story_changeset(attrs)
   end
 
   defp maybe_validate_required(changeset, opts, fun) when is_function(fun, 1) do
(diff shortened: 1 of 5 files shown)


lib/t/accounts/profile.ex
```diff
@@ -92,11 +98,125 @@ defmodule T.Accounts.Profile do
...
+  end
+
+  defp validate_page(page), do: {page, _no_labels_no_errors = []}
+
+  @contacts ["telegram", "instagram", "whatsapp", "phone", "email"]
+
+  @spec validate_labels([map], [map], [String.t()]) :: {[map], [String.t()]}
+  defp validate_labels([label | labels], validated, errors) do
+    case label do
+      %{"question" => q} when q in @contacts ->
```
ruslandoga, line 138: only contact labels are validated right now

lib/t/accounts/profile.ex
```diff
@@ -92,11 +98,125 @@ defmodule T.Accounts.Profile do
...
+        case validate_contact(q, label) do
+          %Ecto.Changeset{valid?: true} = cs ->
+            validate_labels(labels, [apply_changes(cs) | validated], errors)
+
+          %Ecto.Changeset{valid?: false} = cs ->
+            validate_labels(labels, validated, render_contact_errors(cs) ++ errors)
+        end
+
+      label ->
+        validate_labels(labels, [label | validated], errors)
```
ruslandoga, line 148: other label validation can be put here in the future

lib/t/accounts/profile.ex
```diff
@@ -92,11 +98,125 @@ defmodule T.Accounts.Profile do
...
+    validate_story(pages, [page | validated_story], new_errors ++ errors)
+  end
+
+  defp validate_story([], validated, _no_errors = []), do: {:ok, :lists.reverse(validated)}
+  defp validate_story([], _validated_story, errors), do: {:error, errors}
+
+  @spec validate_page(map) :: {map, [String.t()]}
+  defp validate_page(%{"labels" => labels} = page) do
+    {labels, errors} = validate_labels(labels || [], _validated = [], _errors = [])
+    {%{page | "labels" => labels}, errors}
```
ruslandoga, line 128: only labels are validated right now, other page validation can be put here in the future

priv/gettext/errors.pot
```diff
@@ -93,3 +93,48 @@ msgstr ""
 
 msgid "must be equal to %{number}"
 msgstr ""
+
+#, elixir-autogen, elixir-format
+#: lib/t/accounts/profile.ex:279
+msgid "%{label} can't be blank"
```
ruslandoga, line 99: once ios starts using [error response](https://github.com/getsince/test5/blob/a6dbeeba17126034d52e6377b929fdd88bf0877a/Since/StoryEdit/StoryEditCoordinator.swift#L144-L150), these would need to be translated

akhrail1996 (approved):
👍

PR #611, opened 2022-02-25
add shared ig contacts to private pages

ruslandoga:
This probably shouldn't be merged but rather ran once.

### DIFF
--- a/dev/dev.ex
+++ b/dev/dev.ex
@@ -46,4 +46,101 @@ defmodule Dev do
       end
     end
   end
+
+  @ig_contacts [
+    %{ig: ["@durov"], user_id: "0000017c-0000-0000-0000-ac1100040000"}
+  ]
+
+  import Ecto.Query
+
+  # use this function to get a list like above
+  def ig_contacts do
+    T.Matches.Interaction
+    |> where([c], fragment("data ->> 'type'") == "contact_offer")
+    |> group_by([c], c.from_user_id)
+    |> select([c], %{
+      user_id: c.from_user_id,
+      contacts: fragment("array_agg(data -> 'contacts')")
+    })
+    |> T.Repo.all()
+    |> Enum.map(fn %{contacts: contacts} = info ->
+      %{info | contacts: group_contacts(contacts)}
+    end)
+    |> Enum.map(fn %{user_id: user_id, contacts: contacts} ->
+      %{user_id: user_id, ig: contacts["instagram"]}
+    end)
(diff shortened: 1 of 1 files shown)


dev/dev.ex
```diff
@@ -46,4 +46,340 @@ defmodule Dev do
       end
     end
   end
+
+  @ig_contacts [
```
ruslandoga, line 50: you'd need to regenerate this list and edit it manually since when you run it, it'd likely be different

dev/dev.ex
```diff
@@ -46,4 +46,340 @@ defmodule Dev do
...
+
+      if has_contacts?(story, ig) do
+        acc
+      else
+        [%{user_id: user_id, story: add_private_page(story, ig)} | acc]
+      end
+    end)
+  end
+
+  def insert_new_stories(stories \\ add_private_pages()) do
```
ruslandoga, line 344: once the list is ready, `Dev.insert_new_stories` would replace stories in db

dev/dev.ex
```diff
@@ -46,4 +46,101 @@ defmodule Dev do
...
     end
   end
+
+  @ig_contacts [
+    %{ig: ["@durov"], user_id: "0000017c-0000-0000-0000-ac1100040000"}
+  ]
+
+  import Ecto.Query
+
+  # use this function to get a list like above
```
akhrail1996, line 56: IO.inspect(contacts, limit: :ifinity)

akhrail1996 (approved):

dev/dev.ex
ruslandoga: sorry, it's :infinity

dev/dev.ex
```diff
@@ -46,4 +46,101 @@ defmodule Dev do
...
+    |> Enum.reject(fn %{ig: ig} -> is_nil(ig) end)
+    |> Enum.map(fn %{ig: igs} = i -> %{i | ig: filter_valid_igs(igs)} end)
+    |> Enum.reject(fn %{ig: ig} -> ig == [] end)
+  end
+
+  defp filter_valid_igs(igs) do
+    Enum.filter(igs, fn ig ->
+      Regex.match?(
+        ~r/^[A-Za-z0-9_](?:(?:[A-Za-z0-9_]|(?:\.(?!\.))){0,99}(?:[A-Za-z0-9_]))$/,
+        ig |> String.replace(["@"], "") |> String.trim()
```
ruslandoga, line 81: this needs to be run before filtering otherwise we get invalid username like `@durov  `

akhrail1996:
done

PR #613, opened 2022-02-25
add contacts loop

ruslandoga:


akhrail1996 (approved):
👍
will merge & deploy after release

PR #617, opened 2022-03-01
add todos

akhrail1996:


lib/t/feeds.ex
```diff
@@ -227,9 +227,13 @@ defmodule T.Feeds do
   end
 
   defp profiles_that_accept_gender_q(query, gender) do
-    join(query, :inner, [p], gp in GenderPreference,
-      on: gp.gender == ^gender and p.user_id == gp.user_id
-    )
+    if gender do
```
akhrail1996, line 230: new users do not have gender yet during TutorialFeed

ruslandoga:
@akhrail1996 shouldn't it have been reviewed?

PR #622, opened 2022-03-10
fix match expiration notification texts

akhrail1996:


lib/t/push_notifications/apns.ex
```diff
@@ -87,11 +87,18 @@ defmodule T.PushNotifications.APNS do
   end
 
   def build_alert_payload("match_about_to_expire" = type, data) do
-    %{"name" => name} = data
+    %{"name" => name, "gender" => gender} = data
```
ruslandoga, line 90: `"gender"` wouldn't exist for jobs scheduled before this version.

lib/t/push_notifications/apns.ex
```diff
@@ -87,11 +87,18 @@ defmodule T.PushNotifications.APNS do
...
+
+    body =
+      case gender do
+        "F" -> dgettext("apns", "Contact her ✨")
+        "M" -> dgettext("apns", "Contact him ✨")
+        "N" -> dgettext("apns", "Contact them ✨")
```
ruslandoga, line 96: I prefer templating pronouns / verbs / etc.

```elixir
body = dgettext("Contact %{pronoun} ✨", pronoun: pronoun(gender))
```
```gettext
defp pronoun("F"), do: dgettext("apns", "her")
defp pronoun("M"), do: dgettext("apns", "him")
defp pronoun("N"), do: dgettext("apns", "them")
```

This approach can be used for all messages.

akhrail1996: gettext doesn't seem to like this approach (I had similar problem when fixing "news" localization):

image

ruslandoga: I missed the domain in my example.

```elixir
body = dgettext("apns", "Contact %{pronoun} ✨", pronoun: pronoun(gender))
```

ruslandoga (approved):

PR #623, opened 2022-03-10
remove interactions continuation

akhrail1996:


lib/t_web/channels/feed_channel.ex
```diff
@@ -83,39 +71,12 @@ defmodule TWeb.FeedChannel do
...
-      user.id
-      |> Matches.list_archived_matches()
-      |> render_matches(screen_width)
-
-    {:reply, {:ok, %{"archived_matches" => archived_matches}}, socket}
-  end
-
-  def handle_in("archive-match", %{"match_id" => match_id}, socket) do
-    Matches.mark_match_archived(match_id, me_id(socket))
-    {:reply, :ok, socket}
```
ruslandoga, line 99: Should this reply with an alert? Or are all users on 6.0.0 already?

lib/t_web/channels/feed_channel.ex
```diff
@@ -251,18 +212,6 @@ defmodule TWeb.FeedChannel do
           screen_width: screen_width,
           expiration_date: expiration_date
         })
-
```
ruslandoga, line 254: Since there is no need for multiple function clauses, I'd move match struct pattern matching inside the function body, it would make it a bit easier to read.

ruslandoga:
```elixir
Enum.map(matches, fn match ->
  %Matches.Match{...} = match
  render_match(%{...})
end)
```

priv/repo/migrations/20220310174843_drop_expired_matches.exs
```diff
@@ -0,0 +1,7 @@
+defmodule T.Repo.Migrations.DropExpiredMatches do
+  use Ecto.Migration
+
+  def change do
+    drop table(:expired_matches)
```
ruslandoga, line 5: I suggest against dropping tables in migrations. Instead the app should be made able to work without them. And after it is, the tables can be dropped out-of-band (direct sql query to postgres), and migrations that created them -- removed.

lib/t/feeds.ex
```diff
@@ -244,7 +236,6 @@ defmodule T.Feeds do
     |> not_liked_profiles_q(user_id)
     |> not_liker_profiles_q(user_id)
     |> not_seen_profiles_q(user_id)
-    |> not_expired_match_with_q(user_id)
```
ruslandoga, line 247: Would this change result in users seeing their old matches?

lib/t/matches.ex
```diff
@@ -407,7 +360,6 @@ defmodule T.Matches do
       end
     end)
     |> delete_likes()
-    |> insert_expired_match()
```
ruslandoga, line 410: This seems more like unmatch. Should this function do something more than unmatching or should that operation be named differently?

lib/t_web/channels/feed_channel.ex
akhrail1996: @ruslandoga this PR will be merged when users are on 6.0.0

lib/t/feeds.ex
akhrail1996: @ruslandoga not really, this was a workaround and not a proper solution since expired matches were deleted from ExpiredMatch table (and correspondingly from this query) anyways upon "expired_match" being "seen"

lib/t/matches.ex
akhrail1996: @ruslandoga you are right: this function should also "bump" seen status (so that expired mates do not see each other for seen_prune period)
thanks for noticing 👍

priv/repo/migrations/20220310174843_drop_expired_matches.exs
akhrail1996: @ruslandoga okay, do I understand it correctly that upon release of the version with this PR I need to:
1) connect to prod db
2) drop these tables
3) create a separate PR removing migrations that created these tables
?

ruslandoga: No, this PR just needs to stop using the expired and archived matches tables.

lib/t/matches.ex
```diff
@@ -417,6 +418,30 @@ defmodule T.Matches do
     end)
   end
 
+  defp bump_seen_dates(multi) do
```
ruslandoga, line 421:
```suggestion
  defp mark_seen(multi) do
```

lib/t_web/channels/feed_channel.ex
ruslandoga: I'd prefer merging it before that

ruslandoga:
@akhrail1996 I think the target branch won't be merged anyway. So probably it's better to close this one as well.

PR #629, opened 2022-03-12
make all env vars in prod optional

ruslandoga:


config/runtime.exs
```diff
@@ -34,9 +34,19 @@ config :t, Oban,
 
 config :ex_aws,
   json_codec: Jason,
+  # TODO switch to US
   region: "eu-north-1"
```
ruslandoga, line 38: switch default ex_aws region to uUS to make it be read from env vars

config/runtime.exs
```diff
@@ -34,9 +34,19 @@ config :t, Oban,
...
+smoke? = !!System.get_env("SMOKE")
+
+if config_env() == :prod and smoke? do
+  config :t, T.Media.Static, disabled?: true
+  config :t, T.Periodics, disabled?: true
+  config :t, Finch, disabled?: true
+end
+
+if config_env() == :prod and not smoke? do
+  # TODO remove
```
ruslandoga, line 49: remove admin_id from config

Issue #351, opened 2021-10-21
setup db read replicas


ruslandoga:
Interacting with a database in `eu-north-1` outside of Europe appears to be slow. One of the solutions to reduce latencies is to use read replicas in other regions.

I see two ways for using read replicas in `test3`:

- ~~keeping connections to both read replicas and master node in every instance of backend: https://hexdocs.pm/ecto/replicas-and-dynamic-repositories.html#content~~

- or using [rpc](http://erlang.org/doc/man/erpc.html) approach similar to https://github.com/superfly/fly_rpc_elixir (also [presented on ElixirConf 2021](https://www.youtube.com/watch?v=IqnZnFpxLjI&list=PLqj39LCvnOWZna91xJ_i44g3rx4Brbpnv&index=29)) where instances of backend are only connected to their region's db and relay write request to nodes in master region `eu-north-1` through `:erpc`. This is probably fine for small scale (up to a dozen geo-distributed nodes), it's not very scalable beyond that, but that's a problem for the distant future.

PR #636, opened 2022-03-16
add rpc

ruslandoga:
closes https://github.com/getsince/test3/issues/351


ruslandoga:
https://www.githubstatus.com actions don't seem to work

Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: test
+test:
+	elixir --name t@10.0.1.1 -S mix test
```
ruslandoga, line 3: tests are run as if it's a node in the primary region (`10.0.0.0/16`)

config/runtime.exs
```diff
@@ -142,10 +145,14 @@ if config_env() == :prod and not smoke? do
...
           polling_interval: :timer.seconds(ec2_polling_interval),
-          regions: ["eu-north-1", "us-east-2"]
+          regions: regions
         ]
       ]
     ]
+
+  # TODO use cidr like in PRIMARY_SUBNET=10.0.0.0/16
+  # export PRIMARY_HOST_PREFIX=10.0.
+  config :e, primary_prefix: System.fetch_env!("PRIMARY_HOST_PREFIX")
```
ruslandoga, line 155: - [ ] note to self: add this to env vars

lib/t/accounts.ex
```diff
@@ -445,6 +506,16 @@ defmodule T.Accounts do
...
 
   def get_user_by_session_token_and_update_version(token, version, context) do
+    primary_rpc(__MODULE__, :local_get_user_by_session_token_and_update_version, [
+      token,
+      version,
+      context
+    ])
+  end
+
+  # TODO split update, only do if old_version != new_version
```
ruslandoga, line 516: - [ ] only `primary_rpc` if `old_version != new_version`

lib/t/accounts.ex
```diff
@@ -580,12 +662,18 @@ defmodule T.Accounts do
     # |> notify_subscribers([:user, :new])
   end
 
-  def update_profile(%Profile{audio_only: audio_only} = profile, attrs, opts \\ []) do
+  # TODO can do changeset stuff before rpc
```
ruslandoga, line 665: - [ ] todo

lib/t/application.ex
```diff
@@ -80,13 +80,24 @@ defmodule T.Application do
 
   defp maybe_oban do
     if repo_url() do
-      {Oban, oban_config()}
+      if T.Cluster.is_primary() do
```
ruslandoga, line 83: oban only runs in primary region

lib/t/application.ex
```diff
@@ -80,13 +80,24 @@ defmodule T.Application do
...
+        nil
+      end
     else
       Logger.warn("not starting oban due to missing repo url info")
       nil
     end
   end
 
+  defp maybe_periodics do
+    if T.Cluster.is_primary() do
```
ruslandoga, line 96: periodics (prunes and stuff) only run in primary region

lib/t/cluster.ex
```diff
@@ -1,6 +1,9 @@
 defmodule T.Cluster do
   @moduledoc false
```
ruslandoga, line 2: - [ ] doctest

lib/t/feeds.ex
```diff
@@ -46,6 +48,7 @@ defmodule T.Feeds do
           String.t() | nil
         ) ::
           {[%FeedProfile{}], String.t()}
+  # TODO remove writes
```
ruslandoga, line 51: @akhrail1996 can writes to DB be removed from `fetch_feed`? Then it would benefit from read replicas.

lib/t/feeds.ex
```diff
@@ -143,13 +146,24 @@ defmodule T.Feeds do
   end
 
   defp empty_feeded_profiles(user_id) do
+    primary_rpc(__MODULE__, :local_empty_feeded_profiles, [user_id])
```
ruslandoga, line 149: Otherwise we pay for the writes in terms of latency here and ...

lib/t/feeds.ex
```diff
@@ -143,13 +146,24 @@ defmodule T.Feeds do
...
+  end
+
+  @doc false
+  def local_empty_feeded_profiles(user_id) do
     FeededProfile |> where(for_user_id: ^user_id) |> Repo.delete_all()
   end
 
   defp mark_profiles_feeded(for_user_id, feed_profiles) do
+    feeded_user_ids = Enum.map(feed_profiles, fn profile -> profile.user_id end)
+    primary_rpc(__MODULE__, :local_mark_profiles_feeded, [for_user_id, feeded_user_ids])
```
ruslandoga, line 159: ... here.

lib/t/accounts.ex
```diff
@@ -796,29 +878,40 @@ defmodule T.Accounts do
...
-    |> where([p], p.last_active <= fragment("now() - interval '1 day'"))
-    |> where([p], p.last_active > fragment("now() - interval '1 day 1 minute'"))
-    |> select([p], p.user_id)
-    |> Repo.all()
-    |> Enum.map(fn user_id ->
-      schedule_complete_onboarding_push(user_id)
-    end)
-  end
+  # TODO unique
+  def local_push_users_to_complete_onboarding do
```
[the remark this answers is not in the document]
akhrail1996: @ruslandoga 
> is this important?

yes

> could you please ensure it doesn't run more than once per user in a multi-node setup? 

Yes, will do. I guess the same would need to be done to match soon_to_expire and expiration pushes?

lib/t/feeds.ex
akhrail1996: @ruslandoga I think so, since these profiled should be filtered out by this: https://github.com/getsince/test3/blob/master/lib/t/feeds.ex#L250

gotta test though, and maybe adjust something on frontend as well

lib/t/accounts.ex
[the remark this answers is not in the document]
ruslandoga: `local_expiration_notify_soon_to_expire` can use unique oban jobs

lib/t/accounts.ex
```diff
@@ -132,16 +146,19 @@ defmodule T.Accounts do
     end
   end
 
-  # TODO in one transaction
+  # TODO in one transaction in primary
```
ruslandoga, line 149: - [ ] todo

lib/t/accounts.ex
```diff
@@ -110,15 +120,19 @@ defmodule T.Accounts do
...
-      Repo.insert!(%Profile{
-        user_id: user.id,
-        last_active: DateTime.truncate(DateTime.utc_now(), :second)
-      })
-
+    profile = primary_rpc(__MODULE__, :local_ensure_has_profile, [user.id])
     %User{user | profile: profile}
   end
 
+  # TODO race condition
```
ruslandoga, line 127: - [ ] todo

lib/t/accounts.ex
ruslandoga, line 516: - [ ] todo

lib/t/feeds.ex
akhrail1996: @ruslandoga actually, no, my bad
In any case batch fetching happens before every profile is seen => we can not just remove this. Maybe store them in memory?

ruslandoga: @akhrail1996 can a cursor be used instead of feeded_profiles?

lib/t/accounts.ex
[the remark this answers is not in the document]
ruslandoga: @akhrail1996 for now I think it's easiest to move this logic to account registration where an oban job is scheduled to send a notification in 24h if the user still doesn't have a story/profile. Same for expirations.

lib/t/feeds.ex
ruslandoga: why distance?

ruslandoga: wouldn't there be a language barrier?

akhrail1996: No, I don't think there should be artificial barriers. Since communication happens outside the language barrier is also not a big problem.
Currently I am using postgis for distance-based sorting, and it's working well. But I am open to switching to other solutions if you switch to sqlite.

lib/t/feeds.ex
akhrail1996: but it's still much better than having one node (because of intra-AWS connectivity), right?

ruslandoga: no, latencies are almost the same for public/private networking

lib/t/feeds.ex
akhrail1996: sure: https://github.com/getsince/test3/pull/637
beware, it's ugly 🙃

lib/t/accounts.ex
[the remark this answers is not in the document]
ruslandoga: @akhrail1996 I'm taking over this part

lib/t/feeds.ex
ruslandoga: feed fetching optimisation would happen in some next pr

lib/t/matches.ex
```diff
@@ -224,11 +224,22 @@ defmodule T.Matches do
...
-        job = DispatchJob.new(%{"type" => "match", "match_id" => match.id})
-        Oban.insert(job)
+        before_expire =
+          DateTime.truncate(now, :second)
+          |> DateTime.add(match_ttl())
+          |> DateTime.add(-2 * 3600)
+
+        jobs = [
+          DispatchJob.new(%{"type" => "match", "match_id" => match.id}),
+          DispatchJob.new(%{"type" => "match_about_to_expire", "match_id" => match.id},
```
ruslandoga, line 237: - [ ] insert `match_about_to_expire` jobs for existing matches

ruslandoga: - [ ] also expire matches in an oban job

ruslandoga:
@akhrail1996 please review

lib/t/matches.ex
```diff
@@ -65,10 +67,15 @@ defmodule T.Matches do
...
+    primary_rpc(__MODULE__, :local_like_user, [by_user_id, user_id])
+  end
+
+  @doc false
+  def local_like_user(by_user_id, user_id) do
     Multi.new()
-    |> mark_liked(by_user_id, user_id)
-    |> bump_likes_count(user_id)
-    |> match_if_mutual(by_user_id, user_id)
+    |> mark_liked_m(by_user_id, user_id)
```
akhrail1996, line 76: what is the rationale behind adding "_m"? 🙇

akhrail1996 (approved):

ruslandoga:
> will all the current image assets stay accessible as before?

what image assets? which urls?

PR #637, opened 2022-03-16
sort feed by distance

akhrail1996:


lib/t/feeds.ex
```diff
@@ -73,29 +73,45 @@ defmodule T.Feeds do
...
+    # most_liked_q(user_id, gender, gender_preferences, feeded)
+    # most_liked =
+    feed_profiles_q(user_id, gender, gender_preferences)
+    # |> maybe_apply_age_filters(min_age, max_age)
+    # |> maybe_apply_distance_filter(location, distance)
+    # |> limit(^most_liked_count)
+    # |> select_merge([p], %{distance: st_distance(^location, p.location) / 1000})
+    # select([p], %{p | distance: st_distance(^location, p.location) / 1000})
+    |> select([p], {p, fragment("st_distance(?, location) / 1000", ^location)})
+    |> order_by(fragment("st_distance(?, location) / 1000", ^location))
```
akhrail1996, line 89: So far I was not able to find a way to use normal order_by in Ecto, avoiding double computation (?)
related: https://stackoverflow.com/questions/58579900/order-by-virtual-computed-field

ruslandoga (commented):
you can use position in `select`: 'order by 1' would order by first selected column

lib/t/feeds.ex
ruslandoga: note that st_distance is not in meters

ruslandoga:
also note that you shouldn't render exact distances since it allows for triangulation

akhrail1996:
@ruslandoga yeah, do you think adding some white noise on backend is a solution?

lib/t/feeds.ex
akhrail1996: right, thanks!

PR #662, opened 2022-03-19
add non-accurate distance to feed profile view

akhrail1996:
supporting https://github.com/getsince/test5/pull/246

### DIFF
--- a/lib/t/feeds.ex
+++ b/lib/t/feeds.ex
@@ -94,9 +94,8 @@ defmodule T.Feeds do
       |> maybe_apply_age_filters(min_age, max_age)
       |> maybe_apply_distance_filter(location, distance)
       |> limit(^most_liked_count)
-      |> select([p], {p, distance_km(^location, p.location)})
+      |> select([p], %{p | distance: distance_km(^location, p.location)})
       |> Repo.all()
-      |> with_distance()
 
     filter_out_ids = Enum.map(most_liked, fn p -> p.user_id end)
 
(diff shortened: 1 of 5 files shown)


lib/t/feeds.ex
```diff
@@ -145,6 +147,20 @@ defmodule T.Feeds do
     end
   end
 
+  defp with_distance(query, location) do
+    if location do
+      select_merge(query, [p], %{
+        distance:
+          fragment(
+            "ST_DistanceSphere(?::geometry, location::geometry)::integer / 1000",
```
akhrail1996, line 155: ::integer + integer division ensures non-accuracy of distance (impossible to triangulate the exact location)

ruslandoga:
@akhrail1996 maybe something like https://github.com/getsince/test3/commit/974162afa4b740590d857fa6a13e08b5eac76e66?

ruslandoga:
[Comments](https://old.reddit.com/r/netsec/comments/1ydao6/how_i_was_able_to_track_the_location_of_any/) and the [article itself](https://blog.includesecurity.com/2014/02/how-i-was-able-to-track-the-location-of-any-tinder-user/) seem to show that even rounding in the filter still gives off the information about the real user location with enough data points.

lib/t/feeds.ex
```diff
@@ -152,7 +152,7 @@ defmodule T.Feeds do
       select_merge(query, [p], %{
         distance:
           fragment(
-            "ST_DistanceSphere(?::geometry, location::geometry)::integer / 1000",
+            "round(ST_DistanceSphere(ST_ReducePrecision(?::geometry, 0.01), ST_ReducePrecision(location::geometry, 0.01)) / 1000)::integer",
```
akhrail1996, line 155: @ruslandoga how about using reduced precision?

akhrail1996:
@ruslandoga it works locally I don't understand why this error occurs here:
image
while data types are correct:
image
https://postgis.net/docs/ST_ReducePrecision.html

ruslandoga (approved):

ruslandoga (approved):

PR #666, opened 2022-03-19
add h3 to profiles

akhrail1996:


lib/t/accounts/profile.ex
```diff
@@ -95,6 +97,12 @@ defmodule T.Accounts.Profile do
     %Geo.Point{coordinates: {lon, lat}, srid: 4326}
   end
 
+  defp prepare_h3(%{location: location} = attrs) do
+    Map.put(attrs, :h3, :h3.from_geo(location.coordinates, 10))
```
ruslandoga, line 101: `location.coordinates` are `lon, lat` (postgis convetion, like `x, y`), `:h3` expects `lat, lon`.

lib/t/accounts/profile.ex
```diff
@@ -95,6 +97,12 @@ defmodule T.Accounts.Profile do
...
+  end
+
+  defp prepare_h3(attrs), do: attrs
```
ruslandoga, line 104: I'd drop `Map.delete(attrs, :h3)` here just in case.

lib/t/feeds/feed_profile.ex
```diff
@@ -10,6 +10,7 @@ defmodule T.Feeds.FeedProfile do
     field :hidden?, :boolean
     field :last_active, :utc_datetime
     field :location, Geo.PostGIS.Geometry
+    field :h3, :decimal
```
ruslandoga, line 13: I thought about it a bit, and I think a [custom ecto type](https://hexdocs.pm/ecto/Ecto.Type.html) with `i64 (bigint)` can be used. Before dumping we'd subtract, before reading we'd add `i64.max`, that'd be equivalent to u64.

mix.lock
```diff
@@ -18,6 +18,7 @@
...
   "ecto_sql": {:hex, :ecto_sql, "3.7.2", "55c60aa3a06168912abf145c6df38b0295c34118c3624cf7a6977cd6ce043081", [:mix], [{:db_connection, "~> 2.2", [hex: :db_connection, repo: "hexpm", optional: false]}, {:ecto, "~> 3.7.0", [hex: :ecto, repo: "hexpm", optional: false]}, {:myxql, "~> 0.4.0 or ~> 0.5.0 or ~> 0.6.0", [hex: :myxql, repo: "hexpm", optional: true]}, {:postgrex, "~> 0.15.0 or ~> 0.16.0 or ~> 1.0", [hex: :postgrex, repo: "hexpm", optional: true]}, {:tds, "~> 2.1.1 or ~> 2.2", [hex: :tds, repo: "hexpm", optional: true]}, {:telemetry, "~> 0.4.0 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "3c218ea62f305dcaef0b915fb56583195e7b91c91dcfb006ba1f669bfacbff2a"},
+  "elixir_cmake": {:hex, :elixir_cmake, "0.7.0", "e270538c21813832187032da32ae166404d5e5062956e64df2f7a2c174088ff8", [:mix], [], "hexpm", "677e69092c3d95fb7a5ef1c9b5ee8b90b6e84bf08a4c3efc73a670affd34ebe4"},
```
ruslandoga, line 21: should probably be removed from the lock file `mix deps.clean --unlock --unused`

priv/repo/migrations/20220319224657_add_h3_to_profiles.exs
```diff
@@ -0,0 +1,27 @@
...
+  alias T.Feeds.FeedProfile
+
+  def change do
+    alter table(:profiles) do
+      add :h3, :decimal
+    end
+
+    flush()
+
+    FeedProfile
```
ruslandoga, line 15: Migrations shouldn't use ecto schemas, since they tend to change, and at some point in the future this migration wouldn't work.

```
"profiles"
|> select([p], {p.user_id, p.location})
|> where([p], not is_nil(p.location))
|> Repo.all()
|> Enum.chunk_by(300)
|> Enum.each(fn chunk ->
  updates = Enum.map(chunk, fn {user_id, {lon, lat}} ->
    [user_id: user_id, h3: :h3.from_geo({lat, lon}, 10)]
  end)
  Repo.insert_all("profiles", updates, on_conflict: {:replace, [:h3]}, conflict_target: :user_id)
end)
```

lib/t/accounts/profile.ex
```diff
@@ -95,6 +97,16 @@ defmodule T.Accounts.Profile do
...
 
+  defp prepare_h3(%{"latitude" => lat, "longitude" => lon} = attrs) do
+    Map.put(attrs, "h3", :h3.from_geo({lat, lon}, 10))
+  end
+
+  defp prepare_h3(%{latitude: lat, longitude: lon} = attrs) do
+    Map.put(attrs, :h3, :h3.from_geo({lat, lon}, 10))
+  end
+
+  defp prepare_h3(attrs), do: Map.delete(attrs, :h3)
```
ruslandoga, line 108:
```elixir
defp prepare_h3(attrs), do: Map.drop(attrs, [:h3, "h3"])
```

lib/t/misc/bigint.ex
```diff
@@ -0,0 +1,16 @@
+defmodule BigInt do
+  use Ecto.Type
+  def type, do: :integer
+
+  def cast(bigint) when is_integer(bigint), do: {:ok, bigint - 9_223_372_036_854_775_807}
```
ruslandoga, line 5: I'd leave cast as no-op, or check for validity of h3 cell id (that it's u64).

lib/t/misc/bigint.ex
```diff
@@ -0,0 +1,16 @@
+defmodule BigInt do
+  use Ecto.Type
+  def type, do: :integer
+
```
ruslandoga, line 4: I'd add a module attribute here to avoid typos in the u64.max:

```elixir
@u64_max 9_223_372_036_854_775_807
```

lib/t/misc/bigint.ex
```diff
@@ -0,0 +1,16 @@
...
+  def cast(bigint) when is_integer(bigint), do: {:ok, bigint - 9_223_372_036_854_775_807}
+
+  # Everything else is a failure though
+  def cast(_), do: :error
+
+  def load(data) do
+    {:ok, data + 9_223_372_036_854_775_807}
+  end
+
+  def dump(bigint) when is_integer(bigint), do: {:ok, bigint}
```
ruslandoga, line 14: and subtract `@i64_max` here

lib/t/misc/bigint.ex
```diff
@@ -0,0 +1,16 @@
+defmodule BigInt do
```
ruslandoga, line 1: `H3CellId`

lib/t/misc/bigint.ex
```diff
@@ -0,0 +1,16 @@
...
+  use Ecto.Type
+  def type, do: :integer
+
+  def cast(bigint) when is_integer(bigint), do: {:ok, bigint - 9_223_372_036_854_775_807}
+
+  # Everything else is a failure though
+  def cast(_), do: :error
+
+  def load(data) do
```
ruslandoga, line 10: `load(cell_id)`

lib/t/misc/bigint.ex
ruslandoga, line 14: `dump(cell_id)`

lib/t/misc/bigint.ex
ruslandoga, line 5: `cast(cell_id)`

priv/repo/migrations/20220319224657_add_h3_to_profiles.exs
```diff
@@ -0,0 +1,31 @@
...
+    "profiles"
+    |> select([p], {p.user_id, p.location})
+    |> where([p], not is_nil(p.location))
+    |> T.Repo.all()
+    |> Enum.chunk_every(300)
+    |> Enum.each(fn chunk ->
+      updates =
+        Enum.map(chunk, fn {user_id, location} ->
+          %{coordinates: {lon, lat}} = location
+          [user_id: user_id, h3: :h3.from_geo({lat, lon}, 10) - 9_223_372_036_854_775_807]
```
ruslandoga, line 22: to make it clear to the reader, I'd move `9_223_372_036_854_775_807` to `@i64_max` module attribute as well.

PR #672, opened 2022-03-21
process location from socket params and save it to db

akhrail1996:


lib/t/accounts.ex
```diff
@@ -114,6 +114,17 @@ defmodule T.Accounts do
     |> Repo.update_all(set: [last_active: time])
   end
 
+  def update_location(user_id, location) do
+    primary_rpc(__MODULE__, :local_update_location, [user_id, location])
```
ruslandoga, line 118: For later: `primary_rpc` only if `new_location != old_location (from read replica)`.

ruslandoga (approved):

PR #674, opened 2022-03-22
add "update app" news story page for user of versions without support of new features

akhrail1996:


ruslandoga (changes requested):
looks more like a todo

ruslandoga:
news already have versioning via id

PR #676, opened 2022-03-22
add distance to likes and unseen matches

akhrail1996:
Continuing #662 

Distance is not added to seen matches because it allows continuous tracking of the same people (location + boolean for whether user logged into Since lately)

### DIFF
--- a/lib/t/matches.ex
+++ b/lib/t/matches.ex
@@ -43,21 +43,32 @@ defmodule T.Matches do
     Phoenix.PubSub.broadcast_from(@pubsub, self(), pubsub_user_topic(user_id), message)
   end
 
-  # - Likes
+  defmacrop distance_km(location1, location2) do
+    quote do
+      fragment(
+        "round(? / 1000)::int",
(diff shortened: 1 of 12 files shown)


akhrail1996:
GitHub actions seem to be down, but locally tests pass

image

lib/t/matches.ex
```diff
@@ -42,21 +43,32 @@ defmodule T.Matches do
...
+    quote do
+      fragment(
+        "round(? / 1000)::int",
+        st_distance_in_meters(unquote(location1), unquote(location2))
+      )
+    end
+  end
 
-  @spec like_user(Ecto.UUID.t(), Ecto.UUID.t()) ::
+  defp default_location, do: %Geo.Point{coordinates: {0, 0}, srid: 4326}
```
ruslandoga, line 55: ?

ruslandoga: @akhrail1996 tests shouldn't affect the code they tests, please find another way

akhrail1996:
@ruslandoga the approach to tests or the general approach (passing location to `like_user` and `list_matches`)?

PR #677, opened 2022-03-23
add distance news story

akhrail1996:


lib/t/news.ex
```diff
@@ -134,6 +134,51 @@ defmodule T.News do
             "size" => [375, 667]
           }
         ]
```
ruslandoga, line 136: prev news story can probably be removed

lib/t/news.ex
```diff
@@ -134,6 +134,51 @@ defmodule T.News do
...
+                    "Теперь в ленте показывается\nпримерное расстояние\nдо пользователя."
+                  ),
+                "position" => [24.0, 238.0],
+                "background_fill" => "#F97EB9"
+              },
+              %{
+                "value" =>
+                  dgettext(
+                    "news",
+                    "Включи автоматическое\nопределение локации,\nчтобы она\nоставалась актуальной 👇"
```
ruslandoga, line 167: reads like a command, while AFAIK it's an option

lib/t/news.ex
```diff
@@ -134,6 +134,51 @@ defmodule T.News do
...
+                  ),
+                "position" => [24.0, 356.0],
+                "background_fill" => "#F97EB9"
+              },
+              %{
+                "action" => "enable_auto_location",
+                "value" => dgettext("news", "Включить"),
```
ruslandoga, line 174: I'd call it `Enable autolocation` or something like that. It's important to be clear.

lib/t/news.ex
```diff
@@ -155,10 +200,14 @@ defmodule T.News do
   @doc false
   def local_mark_seen(user_id, news_story_id) do
```
ruslandoga, line 201: please explain this change

priv/gettext/en/LC_MESSAGES/news.po
```diff
@@ -85,3 +86,23 @@ msgstr "We have important news."
...
+
+#, elixir-autogen, elixir-format
+#: lib/t/news.ex:174
+msgid "Включить"
+msgstr "Enable"
+
+#, elixir-autogen, elixir-format
+#: lib/t/news.ex:150
+msgid "Мы добавили\nактивную локацию."
+msgstr "We added\nactive location."
```
ruslandoga, line 103: active location?

lib/t/news.ex
akhrail1996: the previous version didn't allow for update of `last_id`:
image

image

lib/t/news.ex
```diff
@@ -216,4 +228,12 @@ defmodule T.News do
   defp last_seen_id(user_id) do
     SeenNews |> where(user_id: ^user_id) |> select([n], n.last_id) |> Repo.one()
   end
+
+  defp user_inserted_at(user_id) do
```
ruslandoga, line 232: `user_id` already contains timestamp

akhrail1996: @ruslandoga what could decoding look like?

test/t_web/channels/feed_channel_test.exs
```diff
@@ -125,29 +125,9 @@ defmodule TWeb.FeedChannelTest do
...
-      assert [_first_news_item = %{id: 2, story: story}] = news
-
-      assert [p] = story
-
-      assert %{"background" => %{"color" => _}, "labels" => _, "size" => _} = p
-
-      ref = push(socket, "seen", %{"news_story_id" => 2})
-      assert_reply ref, :ok, _
-
+    test "with no news", %{socket: socket, me: me} do
```
akhrail1996, line 128: @ruslandoga do you think:
a) user with older inserted_at should be factory-created here 
b) news should be tested separately as a module
c) sth else?

lib/t/news.ex
ruslandoga: extract the first 64 bits, that's a unix timestamp in ms, turn it into a datetime using from_unix function

test/t_web/channels/feed_channel_test.exs
ruslandoga: `%User{user | id: old_id}`

akhrail1996:
@ruslandoga if eth is all right, can we please merge & deploy this and I will add tests later today?

lib/t/news.ex
```diff
@@ -134,6 +135,52 @@ defmodule T.News do
             "size" => [375, 667]
           }
         ]
+      },
+      %{
+        id: 2,
+        timestamp: DateTime.new!(~D[2022-03-23], ~T[08:16:00.000], "Etc/UTC"),
```
ruslandoga, line 141: please use `~U` sigil

lib/t/news.ex
```diff
@@ -12,6 +12,7 @@ defmodule T.News do
     [
       %{
         id: 1,
+        timestamp: DateTime.new!(~D[2022-03-02], ~T[20:31:00.000], "Etc/UTC"),
```
ruslandoga, line 15: please use `~U` sigil

lib/t/news.ex
```diff
@@ -145,7 +192,19 @@ defmodule T.News do
   @spec list_news(Ecto.Bigflake.UUID.t()) :: [%{id: pos_integer(), story: [map]}]
   def list_news(user_id) do
     last_seen_id = last_seen_id(user_id) || 0
+
     Enum.filter(news(), fn news_story -> news_story.id > last_seen_id end)
+    |> Enum.filter(fn news_story ->
+      <<unix_binary::binary-size(8), _rest::binary>> = Ecto.Bigflake.UUID.dump!(user_id)
+      unix_int = :binary.decode_unsigned(unix_binary)
+
+      user_inserted_at = DateTime.from_unix!(unix_int, :millisecond)
```
ruslandoga, line 201: please move that outside of the loop

ruslandoga:
@akhrail1996 I can refactor later, you can merge if you want

ruslandoga:
ping me when it needs to be deployed

ruslandoga (approved):

akhrail1996:
> ping me when it needs to be deployed

@ruslandoga pinging :)

PR #685, opened 2022-03-31
add min_version to news, add "new contacts" news story

akhrail1996:
min_version is added for items that cannot be rendered correctly in previous versions

e.g. new contacts will be rendered as `interests` stickers:
image
which is worse than not showing them at all

### DIFF
--- a/lib/t/news.ex
+++ b/lib/t/news.ex
@@ -13,6 +13,7 @@ defmodule T.News do
       %{
         id: 1,
         timestamp: ~U[2022-03-02 20:31:00Z],
+        version: "6.0.0",
         story: [
           %{
             "background" => %{"color" => "#111010"},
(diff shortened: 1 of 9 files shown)


ruslandoga (approved):

lib/t/news.ex
```diff
@@ -201,6 +256,12 @@ defmodule T.News do
         _ -> true
       end
     end)
+    |> Enum.filter(fn news_story ->
+      case Version.compare(news_story.version, version) do
```
ruslandoga, line 260:
```elixir
|> Enum.filter(fn news_story ->
  Version.compare(version , news_story.version) == :gt
end)
```

ruslandoga:
```suggestion
|> Enum.reject(fn news_story ->
  Version.compare(version, news_story.version) == :lt
end)
```

PR #686, opened 2022-04-01
sort feed by distance

akhrail1996:


dev/dev.ex
```diff
@@ -46,4 +46,75 @@ defmodule Dev do
       end
     end
   end
+
+  def onboard_users_with_random_locations(count) do
```
akhrail1996, line 50: used for insertion of users with random location, super ineffective

ruslandoga (approved):

PR #693, opened 2022-04-11
add s3_key to story background with proxy and color but no s3_key

### LINKED ISSUE #691: correctly process story background with s3_key and additional payload + add test

akhrail1996:
in parallel, I will fix the frontend so it submits s3_key instead of a proxy

closes #691

### DIFF
--- a/lib/t_web/channels/profile_channel.ex
+++ b/lib/t_web/channels/profile_channel.ex
@@ -107,11 +107,14 @@ defmodule TWeb.ProfileChannel do
   end
 
   defp replace_photo_url_with_s3key(%{"s3_key" => _} = background), do: background
-  defp replace_photo_url_with_s3key(%{"color" => _} = bg), do: bg
 
-  defp replace_photo_url_with_s3key(%{"proxy" => proxy_url}) do
-    %{"s3_key" => s3_key_from_proxy_url(proxy_url)}
+  defp replace_photo_url_with_s3key(%{"proxy" => proxy_url} = background) do
+    background
+    |> Map.put("s3_key", s3_key_from_proxy_url(proxy_url))
+    |> Map.delete("proxy")
   end
+
+  defp replace_photo_url_with_s3key(%{"color" => _} = bg), do: bg
 
   # TODO fix, client shouldn't be sending urls
   defp s3_key_from_proxy_url("https://d3r9yicn85nax9.cloudfront.net/" <> path) do
--- a/test/t_web/channels/profile_channel_test.exs
+++ b/test/t_web/channels/profile_channel_test.exs
@@ -342,6 +342,48 @@ defmodule TWeb.ProfileChannelTest do
     end
   end
 
+  describe "story update" do
+    setup do
+      user = onboarded_user()
+      {:ok, user: user, socket: connected_socket(user)}
+    end
+
+    setup :subscribe_and_join
+
+    test "background with proxy and position but no s3 key", %{socket: socket} do
+      story = [
+        %{
+          "background" => %{
+            "zoom" => 1.1315270791341807,
+            "color" => "#6B4D32",
+            "proxy" =>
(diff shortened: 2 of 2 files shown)


lib/t_web/channels/profile_channel.ex
```diff
@@ -107,12 +107,15 @@ defmodule TWeb.ProfileChannel do
   end
 
   defp replace_photo_url_with_s3key(%{"s3_key" => _} = background), do: background
-  defp replace_photo_url_with_s3key(%{"color" => _} = bg), do: bg
```
akhrail1996, line 110: the issue was this matching the story background with "proxy" + position (incl. "color") -> not getting s3_key

ruslandoga (changes requested):
I'm missing what exactly this is solving. Can it be solved similar to stickers?

1. fix the client
2. remove `replace_photo_url_with_s3key`
3. story changeset function should validate/restore s3_key for background similar to labels
4. cleanup db

lib/t_web/channels/profile_channel.ex
```diff
@@ -107,12 +107,15 @@ defmodule TWeb.ProfileChannel do
...
 
-  defp replace_photo_url_with_s3key(%{"proxy" => proxy_url}) do
-    %{"s3_key" => s3_key_from_proxy_url(proxy_url)}
+  defp replace_photo_url_with_s3key(%{"proxy" => proxy_url} = background) do
```
ruslandoga, line 111: what's in `background`?

ruslandoga:
> I'm missing what exactly this is solving.

Is the actual change on in the order of the function clauses?

lib/t_web/channels/profile_channel.ex
akhrail1996: e.g. https://github.com/getsince/test3/pull/693/files#diff-4b57c8e03ec7e4456f02252bb310d0189cf01a8285d6e238c9698a5a78c2bc57R357

akhrail1996:
> Can it be done similar to stickers?

yes, will do. good idea, thanks!


> Is the change only in the order of the function clauses? s3_key > color > proxy becomes s3_key > proxy > color?

Also: https://github.com/getsince/test3/pull/693/files#diff-cd42229fa05d0e19bbee9514e6e6d57b07cb357bd02acf85336fb421332f8f41L113

ruslandoga:
@akhrail1996 why zoom background?

akhrail1996:
> @akhrail1996 why zoom background? I thought you were rasterising it on the client. Otherwise imgproxy is useless.

to allow resized (non-full screen) backgrounds:
image

full-screen (non-resized) background still do not have 'zoom'

ruslandoga:
zoom > 1.0 -> crop, don't zoom
zoom < 1.0 -> update imgproxy scales

akhrail1996:
> zoom > 1.0 -> crop, don't zoom

not nessecarily crop, for example:

no zoom (1.0)
image

zoom > 1.0
image

> zoom < 1.0 -> update imgproxy scales when picking [width bucket](https://github.com/getsince/test3/blob/225acee9c08a9d9f01a4ca5c67d9b5a69c43272d/lib/t/media.ex#L353)

👍, https://github.com/getsince/test5/issues/264

ruslandoga:
@akhrail1996 are prev uploads assumed to be zoom = 1.0, does it mean the photos int he stories are not `fit` instead of `filled`?

and how does it work right now? If s3_key is not saved, what do the users see, proxy or color?

akhrail1996:
> @akhrail1996 are prev uploads assumed to be zoom = 1.0

no, these are separate. all the prev uploads are ".scaleAspectFill"

> does it mean the photos int he stories are not `fit` instead of `filled`?

all the current uploads are '.scaleAspectFit'

> and how does it work right now? If s3_key is not saved, what do the users see, proxy or color?

proxy

ruslandoga:
> all the current uploads are '.scaleAspectFit', regardless of zoom

how are they separated from the prev ones?

akhrail1996:
> how are they separated from the prev ones?

prev ones do not have 'color' property (average color of the image)

PR #700, opened 2022-04-16
drop support of v5, remove text-contact, bump minor version

akhrail1996:


dev/text_contact.ex
```diff
@@ -22,7 +22,29 @@ defmodule Dev.StoryBackground do
...
   defp fix_profiles(profiles) do
     Enum.map(profiles, fn %{story: story} = profile ->
-      %{profile | story: StoryBackground.fix_story(story)}
+      %{profile | story: fix_story(story)}
+    end)
+  end
+
+  def fix_story(nil), do: nil
+
+  def fix_story(story) when is_list(story) do
```
ruslandoga, line 31: @akhrail1996 this needs to be run first, then the update to views can be deployed

PR #701, opened 2022-04-18
add Spotify token fetch

akhrail1996:
enables users to make basic API requests which do not requite user auth (search for tracks)

### DIFF
--- a/lib/t/spotify.ex
+++ b/lib/t/spotify.ex
@@ -1,22 +1,75 @@
 defmodule T.Spotify do
   @moduledoc "spotify api client"
 
+  use GenServer
+
   @spotify_token_url "https://accounts.spotify.com/api/token"
 
-  def token do
-    config = Application.get_env(:t, __MODULE__)
-    token(config)
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
   end
 
-  # TODO cache?
-  def token(config) do
+  @spec current_token :: {:ok, String.t()} | :error
+  def current_token do
+    token = lookup_token() || refresh_token()
+
+    case token do
+      {token, _expiration_time} -> {:ok, token}
+      _ -> :error
+    end
+  end
+
+  @spec lookup_token :: {String.t(), integer()} | nil
+  defp lookup_token do
+    case :ets.lookup(__MODULE__, :token) do
+      [{_, token}] -> if token_still_valid(token), do: token
+      [] -> nil
+    end
+  end
+
+  defp find_spotify_key do
+    config = Application.fetch_env!(:t, T.Spotify)
     client_id = Keyword.fetch!(config, :client_id)
     client_secret = Keyword.fetch!(config, :client_secret)
 
+    %{client_id: client_id, client_secret: client_secret}
+  end
+
+  @spec refresh_token :: String.t()
+  defp refresh_token do
+    key = find_spotify_key()
+    GenServer.call(__MODULE__, {:refresh_token, key})
+  end
+
+  @impl true
+  def init(_opts) do
+    __MODULE__ = :ets.new(__MODULE__, [:named_table])
+    {:ok, nil}
+  end
+
+  @impl true
+  def handle_call({:refresh_token, key}, _from, state) do
+    %{client_id: client_id, client_secret: client_secret} = key
+
+    if token = lookup_token() do
+      {:reply, token, state}
+    else
+      token = request_token(client_id, client_secret)
+      :ets.insert(__MODULE__, {:token, token})
+      {:reply, token, state}
+    end
+  end
+
+  defp token_still_valid({_token, expiration_time}, now \\ :os.system_time(:second)) do
+    expiration_time > now
+  end
+
+  @spec request_token(String.t(), String.t(), integer) :: {String.t(), integer()} | nil
+  defp request_token(client_id, client_secret, now \\ :os.system_time(:second)) do
     auth_key = Base.encode64(client_id <> ":" <> client_secret)
 
     headers = [
-      {"authorization", "basic " <> auth_key},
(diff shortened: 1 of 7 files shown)


akhrail1996:
@ruslandoga do you think that a test should be added to this PR?

lib/t/spotify.ex
```diff
@@ -0,0 +1,40 @@
...
+  @moduledoc "spotify api client"
+
+  @spotify_token_url "https://accounts.spotify.com/api/token"
+
+  def token do
+    config = Application.get_env(:t, __MODULE__)
+    token(config)
+  end
+
+  # TODO cache?
```
ruslandoga, line 11: cache

lib/t/spotify.ex
```diff
@@ -0,0 +1,40 @@
...
+  def token(config) do
+    client_id = Keyword.fetch!(config, :client_id)
+    client_secret = Keyword.fetch!(config, :client_secret)
+
+    auth_key = Base.encode64(client_id <> ":" <> client_secret)
+
+    headers = [
+      {"authorization", "basic " <> auth_key},
```
ruslandoga, line 19: Basic

lib/t/spotify.ex
```diff
@@ -0,0 +1,40 @@
...
+      {"content-type", "application/x-www-form-urlencoded"}
+    ]
+
+    body = "grant_type=client_credentials"
```
ruslandoga, line 23: ?

lib/t/spotify.ex
```diff
@@ -0,0 +1,40 @@
...
+      {:ok, %Finch.Response{status: _status, body: body, headers: _headers}} ->
+        case Jason.decode(body) do
+          {:ok, %{"access_token" => token}} -> {:ok, %{token: token}}
+          {:ok, %{"error" => error}} -> {:error, %{reason: error}}
+          {:ok, body} -> {:error, %{reason: body}}
+          {:error, error} -> {:error, %{reason: error}}
+        end
+
+      {:error, reason} ->
+        {:error, %{reason: reason}}
```
ruslandoga, line 37: ?

lib/t_web/channels/profile_channel.ex
```diff
@@ -71,6 +71,13 @@ defmodule TWeb.ProfileChannel do
     {:reply, reply, socket}
   end
 
+  # TODO refresh after one hour
```
ruslandoga, line 74: why?

lib/t_web/channels/profile_channel.ex
```diff
@@ -71,6 +71,13 @@ defmodule TWeb.ProfileChannel do
...
+  def handle_in("get-spotify-token", _params, socket) do
+    token = socket.assigns[:spotify_token] || Spotify.token()
+    socket = assign(socket, spotify_token: token)
```
ruslandoga, line 77: why is caching per user?

ruslandoga:
@akhrail1996 what would you test?

lib/t/spotify.ex
ruslandoga: the caller doesn't seem to handle the error in any way, also why wrap the error reason in a map?

ruslandoga:
@akhrail1996 did you copy it from somewhere, did we have spotify before?

lib/t/spotify.ex
akhrail1996: required

image

from https://developer.spotify.com/documentation/general/guides/authorization/client-credentials/

lib/t_web/channels/profile_channel.ex
```diff
@@ -71,6 +71,13 @@ defmodule TWeb.ProfileChannel do
     {:reply, reply, socket}
   end
 
+  # TODO refresh after one hour
+  def handle_in("get-spotify-token", _params, socket) do
+    token = socket.assigns[:spotify_token] || Spotify.token()
```
ruslandoga, line 76: should it be `{:ok, token}`?

lib/t/spotify.ex
ruslandoga: but why in body?

lib/t_web/channels/profile_channel.ex
akhrail1996: token expires in 3600 based on the reply to this request

lib/t_web/channels/profile_channel.ex
akhrail1996: right, no need to

akhrail1996:
> @akhrail1996 did you copy it from somewhere, did we have Spotify before?

no, we had Apple Music and I was partly basing on it

lib/t_web/channels/profile_channel.ex
ruslandoga: check how apns token works in apns/token.ex

akhrail1996:
> @akhrail1996 what would you test?

Maybe just that it gets the token, but that will require adding env variables

Which afaik we cannot do because either GitHub actions won't work or we will reveal our Spotify creds. Though maybe there is a way to be able to add this test for local execution and for Github Actions to skip it?

ruslandoga:
> Maybe just that it gets the token

gets the token from spotify or from a mock?

lib/t/spotify.ex
akhrail1996: I am a novice at this, where should it be?

ruslandoga:
> afaik we cannot do because either GitHub actions won't work

why wouldn't they work?

ruslandoga:
> we will reveal our Spotify creds

do they provide sandboxing?

ruslandoga:
> Though maybe there is a way to be able to add this test for local execution and for Github Actions to skip it?

search for tags and skipping tests in ex_unit

ruslandoga (approved):

PR #707, opened 2022-04-30
add video stickers

akhrail1996:


lib/t_web/views/view_helpers.ex
```diff
@@ -69,7 +69,19 @@ defmodule TWeb.ViewHelpers do
           |> Enum.reduce([], fn label, acc ->
             case label do
               %{"s3_key" => key, "question" => "audio"} ->
-                label = Map.put(label, "url", Accounts.media_url(key))
+                label = Map.put(label, "url", media_cdn_url(key))
+                [label | acc]
+
+              %{
+                "s3_key" => key,
+                "placeholder_s3_key" => placeholder_s3_key,
```
ruslandoga, line 77: `placeholder_s3_key` -> `s3_key`, `s3_key` -> `video_s3_key` to be consistent with video backgrounds?

lib/t_web/views/view_helpers.ex
```diff
@@ -51,7 +50,7 @@ defmodule TWeb.ViewHelpers do
     end
   end
 
-  defp process_labels(%{"labels" => labels} = page, version) do
+  defp process_labels(%{"labels" => labels} = page, version, screen_width) do
     # we do not add url for versions prior to 6.2.0 because they will be rendered incorrectly
     labels =
       case Version.compare(version, "6.2.0") do
```
ruslandoga, line 56: needs another version check, maybe against v6.3.0?

ruslandoga (approved):

lib/t_web/views/view_helpers.ex
akhrail1996: right

lib/t_web/views/view_helpers.ex
akhrail1996: may do it, but not sure if I should since the "video" (unknown question) is "rendered" to nil in 6.2.0

PR #747, opened 2022-07-22
update runtime and ci

ruslandoga:


.github/workflows/build.yml
```diff
@@ -12,36 +12,36 @@ env:
 jobs:
   docker:
     name: docker build amd64
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
ruslandoga, line 15: ensures same os version as in aws

.github/workflows/release.yml
```diff
@@ -23,33 +23,13 @@ jobs:
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
     steps:
-      - uses: actions/checkout@v2
-
-      # TODO use setup-beam once https://github.com/erlef/setup-beam/pull/124 is released
```
ruslandoga, line 28: it's released

mix.exs
```diff
@@ -5,7 +5,7 @@ defmodule T.MixProject do
     [
       app: :t,
       version: "0.1.6",
-      elixir: "~> 1.13",
+      elixir: "~> 1.14",
```
ruslandoga, line 8: might be needed to use brew locally to update elixir to 1.14

```sh
> brew update && brew upgrade elixir erlang
```

.github/workflows/build.yml
akhrail1996: will it become unsupported in 3 months or so?

ruslandoga: it's an LTS (long term support release) and it'd be supported for 6 years iirc

PR #783, opened 2022-09-22
add extra metadata to sentry feed ai reports

ruslandoga:


lib/t/feed_ai.ex
```diff
@@ -250,12 +250,19 @@ defmodule T.FeedAI do
 
   @doc false
   def ssh_into_retry(%{private_ip: ip} = ec2) do
+    Logger.metadata(ec2: ip, host: own_host())
```
ruslandoga, line 253: I'll need to check if that's enough or there's extra sentry config needed.

